PR #1, opened 2020-12-23
Add al_general.py

purplesky2016:


purplesky2016:
I added the py file and versioned it up. 

I don't understand what the change in setup.py (install_requires=[...]) is about. Is it because I used 3 digits in the version instead of 2?

PR #16, opened 2021-01-28
Ask address of the user to help find the court

BryceStevenWilley:
https://github.com/SuffolkLITLab/docassemble-assemblylinewizard/issues/121#issuecomment-769230682 has most of the background for this change. 

We should consider that this will **force the user to give their address**, even if the form doesn't need it, which seems iffy. Some other (more complicated) ways of fixing this missing address issue:
* have a different address ask screen that doesn't mark the address as required, but instead indicates that entering their address will help them later in the interview but isn't necessary for the form itself (do courts always need the address of the user? We might eventually want it when we gather interview stats)
* Add an intermediate `id: choose a court (no address given)` screen that looks like `id:choose a court (no matching courts found)`, but also asks for an optional address, and if the address is entered, runs another search, and then goes to either `(no matching courts found)` or `(courts matching provided address were found)` screens.
* Just mark the linked issue above as Not a Bug, and force interview writers to add `users[0].address` before any questions that trigger court location questions if they want the court selector to be narrowed.

### DIFF
--- a/docassemble/AssemblyLine/data/questions/ql_baseline.yml
+++ b/docassemble/AssemblyLine/data/questions/ql_baseline.yml
@@ -674,6 +674,8 @@ comment: |
   the address where the incident took place. Especially
   if this is a responsive case.
 code: |
+  # Force the address question to be asked if it hasn't been asked yet
+  users[0].address
   addresses_to_search = [users[0].address]
 ---
 comment: |
--- a/docassemble/AssemblyLine/data/questions/ql_baseline.yml
+++ b/docassemble/AssemblyLine/data/questions/ql_baseline.yml
@@ -675,7 +675,7 @@ comment: |
   if this is a responsive case.
 code: |
   # Force the address question to be asked if it hasn't been asked yet
-  users[0].address
+  users[0].address.address
   addresses_to_search = [users[0].address]
 ---
 comment: |


nonprofittechy:
I'm definitely leaning towards letting the interview author decide if they want to ask for the user's address. But given that approach led to this series of bug reports, maybe the how and why needs to spelled out better. Thoughts?

It's very possible to file a form that lists a court but does not ask for the user's address. We need it if the form is being electronically filed, but if it's going to be downloaded, we really don't.

The optional address prompt is a nice idea.

BryceStevenWilley:
Yeah, better documentation on how some of this works would be a good idea. We'd need to figure out where in https://suffolklitlab.org/docassemble-AssemblyLine-documentation/docs/ it should go, and it's probably a good idea to describe quirks of docassemble like that anyway. 

We don't ask if they want to email until the very end though, right? We'd have to move up that question further, which I don't think would be too bad.

I like the optional prompt too, but it might be confusing for the user if we show the optional prompt to them, and then show the a required prompt because an author didn't organize their interview well.

Also realizing a fourth option: make sure we sort the built-in questions correctly in the weaver: always asking user info like address and name before court, and other dependencies that we could figure out. I'm leaning towards this option actually, since it's still flexible, but gets the default use case for authors right.

nonprofittechy:
> 
> 
> Also realizing a fourth option: make sure we sort the built-in questions correctly in the weaver: always asking user info like address and name before court, and other dependencies that we could figure out. I'm leaning towards this option actually, since it's still flexible, but gets the default use case for authors right.

I like this solution a lot, especially since the user doesn't get to rearrange the order of those questions in the Interview Weaver. We don't have consistent style guide choice about that yet. On some forms, we ask for things like the user's name at the beginning of the interview, and some at the end.

I think there was some disagreement about the right choice. Some people felt that it was important to speak to the user throughout the interview using their first name. Others that it was important to allow the user to see the value of the interview before asking for personal information, to build trust.

nonprofittechy:
So I just noticed that the current wizard generates a code block that requires the definition of the user's zip code. So I think we can make address mandatory.

PR #23, opened 2021-02-10
Added generic attorney ID question

### LINKED ISSUE #18: Add a question for bbo
This is Massachusetts specific, but we're okay with that for now. In the future--link this to MA-specific version of the wizard.

We could make the MassAccess package have a list of reserved keywords as a way to make this more flexible in the future. The wizard could choose the organization repo and then select the file which has the right list of reserved keywords in it.

BryceStevenWilley:
Partially closes #18, which will be fully closed when the BBO number question is merged in ALMassachusetts.

If ALMassachusetts isn't included in an interview, this ask for a generic attorney ID number, with no validation.

### DIFF
--- a/docassemble/AssemblyLine/data/questions/ql_baseline.yml
+++ b/docassemble/AssemblyLine/data/questions/ql_baseline.yml
@@ -516,6 +516,19 @@ fields:
   - Email: x.email
     datatype: email
 ---
+id: attorney id number
+question: |
+  What is ${ attorneys[i].possessive('ID number') }?
+fields:
+  - BBO: attorneys[i].id_number
+help: 
+  label: |
+    ID?
+  content: |
+    In some jurisdictions, attorneys will have an identification number. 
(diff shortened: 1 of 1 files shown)


nonprofittechy (approved):
LGTM. I'm okay with .id_number. We should consider making .bbo a synonym on both sides--add a code block that sets attorneys[i].bbo=id_number, and also accepts attorney_bbo

BryceStevenWilley:
That's a good point. Just to clarify, we'd want something like:
```yaml
---
code: |
  attorney[i].bbo = attorney[i].id_number
---
code: |
  attorney[i].id_number = attorney_bbo
---
```
in AssemblyLine, or in ALMassachusetts?

nonprofittechy:
We could use it in ALMassachusetts. And we only need the first code block, not the second one--that way the question you wrote will be used for both BBO and id_number

```
---
code: |
  attorney[i].bbo = attorney[i].id_number
```

PR #41, opened 2021-03-03
Close #40, and invisible sought var data to the dev DOM

### LINKED ISSUE #40: Add name of sought variable to some part of the DOM in the development environment
Somewhere on the developer HUD seems most appropriate.

plocket:
Close #40. You can test with [al_question_test.yml](https://apps-dev.suffolklitlab.org/interview?i=docassemble.playground12ALSoughtVar%3Aal_question_test.yml#page1) for a simple variable name and with [a test I made](https://apps-dev.suffolklitlab.org/interview?i=docassemble.playground12ALSoughtVar%3Atest.yml#page1) for a more complex variable name. It doesn't matter much as the variable name gets encoded.

### DIFF
--- a/docassemble/AssemblyLine/data/questions/al_visual.yml
+++ b/docassemble/AssemblyLine/data/questions/al_visual.yml
@@ -18,6 +18,7 @@ default screen parts:
     % if get_config('debug'):
     `id: ${ user_info().question_id }`  
     `Package: ${ user_info().package } ${ package_version_number }; ${ al_version }`
+    <div data-variable="${ encode_name(user_info().variable) }" id="sought_variable" aria-hidden="true" style="display: none;"></div>
     % endif
 ---
 features:
--- a/docassemble/AssemblyLine/data/questions/al_visual.yml
+++ b/docassemble/AssemblyLine/data/questions/al_visual.yml
@@ -18,7 +18,7 @@ default screen parts:
     % if get_config('debug'):
     `id: ${ user_info().question_id }`  
     `Package: ${ user_info().package } ${ package_version_number }; ${ al_version }`
-    <div data-variable="${ encode_name(user_info().variable) }" id="sought_variable" aria-hidden="true" style="display: none;"></div>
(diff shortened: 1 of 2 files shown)


docassemble/AssemblyLine/data/questions/al_visual.yml
```diff
@@ -18,6 +18,7 @@ default screen parts:
     % if get_config('debug'):
     `id: ${ user_info().question_id }`  
     `Package: ${ user_info().package } ${ package_version_number }; ${ al_version }`
+    <div data-variable="${ encode_name(user_info().variable) }" id="sought_variable" aria-hidden="true" style="display: none;"></div>
```
BryceStevenWilley, line 21: I don't think `encode_name` handles `None` well, and `user_info().variable` is None if there's no variable being sought (purely informational screens? events?). Trying to figure out how to trigger it, but I think this will need to be caught and handled here.

https://github.com/jhpyle/docassemble/blob/10507a53d293c30ff05efcca6fa25f6d0ded0c93/docassemble_base/docassemble/base/functions.py#L4470

https://docassemble.org/docs/functions.html#user_info

nonprofittechy: What about just `encode_name(str(user_info().variable)))`

plocket: Thanks for catching this! The last of those suggestions is one I can implement, thanks!

plocket: Hmm, even with an event screen, it looks like there's a variable being sought. Even so, better safe than sorry.

nonprofittechy: I think the `None` would be triggered if it's a `mandatory` screen that isn't linked to a variable name. I have seen it before in our interviews though when testing the Github feedback form, so there must be another way to trigger it too.

BryceStevenWilley (approved):

PR #46, opened 2021-03-09
ALDocumentizing Sending to Court code

BryceStevenWilley:
Uses the `al_court_bundle` and the `al_user_bundle` `ALDocumentBundle` objects to generate the final document that is sent to the court. Also adds a `get_titles` to the bundle object to get just the titles of all of the forms (which is added to the cover sheet).

The Weaver (as of https://github.com/SuffolkLITLab/docassemble-assemblylinewizard/issues/139) should generate interviews with the `al_court_bundle` and `al_user_bundle`, which is all interviews generated with Weaver 0.55 or later (https://github.com/SuffolkLITLab/docassemble-assemblylinewizard/commit/6577b917beab99d8ce346f0a26cb3b7ea2731d35). Not sure if anyone made a doc between the official switch to AssemblyLine away from MAVirtualCourt and then, but seems small enough for us to find and fix?

The blast radius of this PR seems larger than it should be, any advice is helpful.


nonprofittechy (changes requested):
I would actually expand the scope to replace `form_to_file_no_cover` etc. The send to court code can use the `al_court_bundle` variable directly.

nonprofittechy:
Looking close--LMK when it's ready for a review

docassemble/AssemblyLine/data/questions/al_code.yml
```diff
@@ -74,20 +74,20 @@ objects:
 reconsider: True
 code: |
   al_court_bundle.insert(0, al_cover_page)
-  final_form_to_be_filed = al_court_bundle.as_pdf('final') 
+  al_final_form_to_file = al_court_bundle.as_pdf('final')
```
nonprofittechy, line 77: In most cases the court will want separate documents, not one big PDF. At least this is true for appeals court.

BryceStevenWilley: Like all of the documents separately attached to the email? That's doable I guess, I'd just been avoiding changing as much outward-facing features as possible.

BryceStevenWilley:
Okay, should be actually ready for a final review @nonprofittechy. The only thing I wasn't sure about was https://github.com/SuffolkLITLab/docassemble-AssemblyLine/pull/46/files#diff-55dff7108dded7bbc3ce8459dd669a79baec6674c1558af86f66dbc878268288R958: should we show the users the exact thing the clerk will see? i.e. another list of individual forms? Or is it okay to just shove into a single PDF?

It's been tested on a local docassemble host, in the Motion to Stay General interview and https://github.com/SuffolkLITLab/docassemble-MassAppealsCourt/pull/3 (plus those changes to the Motion to Stay General in that PR).

nonprofittechy:
I think for users the exact thing clerks see is best as a single pdf.
Thanks!

Yeah realizing that the only place we had been sending multiple files was
for the appeals court and it was hardcoded. But that is the most important
ongoing partnership

BryceStevenWilley:
Okay, this has been tested with the right branch this time: https://github.com/SuffolkLITLab/docassemble-MotionToStayGeneral/tree/Changing-to_ALDocument. Should be good.

PR #47, opened 2021-03-10
Name fields with business option

BryceStevenWilley:
Refactored the existing code a bit to also allow people to ask for `show if`'d code for business or people. Existing question structure was taken from the `other_parties` default question, but needed it for `users` as well for some changes in the [Civil docketing statement](https://github.com/SuffolkLITLab/docassemble-AppealsCivilDocketingStatement/issues/12).


nonprofittechy (changes requested):
This looks good. New keyword parameter is fine.

Only substantive change I would make is to use radio buttons, not drop-down, as drop-down takes more clicks and isn't consistent with our style guide (drop down with > 7 choices or infrequently interacted with element)

BryceStevenWilley:
Thanks for the suggestion, had meant to use a radio like the original question, but just forgot to get it in

Issue #29, opened 2021-02-17
Style: improve the look of the read aloud / "playback" button


nonprofittechy:
There are some examples of more unobtrusive widgets that do this around the web. We should be able to style this better.

MLRI's is one good example: 
![image](https://user-images.githubusercontent.com/7645641/108247635-e7654d00-7120-11eb-9999-c976ef3d3e46.png)

What I like about this is that it has a label and icon that more clearly conveys what it does and that the button does not take the full width of the CSS box.

nonprofittechy:
@purplesky2016 has some experience trying to customize the MassLegalHelp version of this input.

@CaroRob might have the model website she liked better

nonprofittechy:
This can be started now--it will be useful. Or it can wait for an interested designer volunteer.

purplesky2016:
I made an attempt, see https://github.com/SuffolkLITLab/docassemble-AssemblyLine/pull/43.

nonprofittechy:
This definitely looks better! @purplesky2016 do you think we can add in a text label to the input with CSS as well?

purplesky2016:
Could you elaborate it a bit?

BryceStevenWilley:
Do you mean something like adding the words "Read the screen" in front of the play button @nonprofittechy? I do think that would be a little better.

purplesky2016:
If we are talking about labels like the following for those radio buttons
![image](https://user-images.githubusercontent.com/20231649/110151103-0d8c1d80-7dae-11eb-985e-f3fc629409cb.png)
we need to be able to control the "input" buttons. It appears that the "audio" button is not directly under our control out of box.

purplesky2016:
This is what it looks like in MS Edge:
![image](https://user-images.githubusercontent.com/20231649/110151979-1fba8b80-7daf-11eb-8581-530f725d9f37.png)

plocket:
I'm going to give a quick [hybrid js/css solution](https://stackoverflow.com/a/39899797/14144258) a shot and see what you folks think. It's looking like it's the standard way people out there handle better looking audio UI.

plocket:
We're choosing to leave out the option of rewinding/scanning back manually, correct? I'm using this as the basic inspiration: https://www.readspeaker.com/

plocket:
Needs testing in different browsers: https://github.com/SuffolkLITLab/docassemble-AssemblyLine/tree/29_custom_audio_2

[...]

PR #52, opened 2021-03-10
#29 audio UI improvements (MVP)

plocket:
Addresses #29, but not completely, only as an MVP. If this version actually makes things worse, we can just close this PR and wait till further development can happen. See #29 for a lot more discussion.


nonprofittechy:
This looks great--thanks Michelle and Lily both

PR #78, opened 2021-04-07
Fix #68, custom audio elem gone when page help button is pressed

### LINKED ISSUE #68: Playback new UI does not show on a "help" page
image

The unstyled HTML5 audio element displays on the help screen, even if the correct new UI appears on other pages of the interview.

plocket:
Fixes #68. For the page event, I would just call it `audio_test`, but I think this may be expanded. For example, it may also be used to test the 'escape' button, etc.

Test performed:
- [ ] Open test interview.
- [ ] See correct custom audio element
- [ ] Test behavior of audio element
- [ ] Tap 'Help'
- [ ] Test behavior of audio element

### DIFF
--- a/docassemble/AssemblyLine/data/static/al_audio.js
+++ b/docassemble/AssemblyLine/data/static/al_audio.js
@@ -2,10 +2,7 @@ $(document).on('daPageLoad', function() {
   var $audio_nodes = $('.daaudio-control');
   var id_count = 1;
   for ( var audio_node of $audio_nodes ) {
-    (function () {
-      console.log( audio_node.constructor.name );
-      al_js.replace_with_audio_minimal_controls( audio_node, 'page_reader_substitute_' + id_count );
-    })();
+    al_js.replace_with_audio_minimal_controls( audio_node, 'page_reader_substitute_' + id_count );
     id_count++;
   }
   
--- a/docassemble/AssemblyLine/data/static/al_audio.js
+++ b/docassemble/AssemblyLine/data/static/al_audio.js
@@ -1,5 +1,14 @@
 $(document).on('daPageLoad', function() {
-  al_js.audio_minimal_controls( $('.daaudio-control')[0], 'page_reader_substitute' );
+  var $audio_nodes = $('.daaudio-control');
+  var id_count = 1;
+  for ( var audio_node of $audio_nodes ) {
+    (function () {
+      console.log( audio_node.constructor.name );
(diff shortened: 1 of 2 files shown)


docassemble/AssemblyLine/data/static/al_audio.js
```diff
@@ -1,5 +1,14 @@
 $(document).on('daPageLoad', function() {
-  al_js.audio_minimal_controls( $('.daaudio-control')[0], 'page_reader_substitute' );
+  var $audio_nodes = $('.daaudio-control');
+  var id_count = 1;
+  for ( var audio_node of $audio_nodes ) {
+    (function () {
+      console.log( audio_node.constructor.name );
+      al_js.replace_with_audio_minimal_controls( audio_node, 'page_reader_substitute_' + id_count );
+    })();
```
BryceStevenWilley, line 8: minor: I've sometimes seen this "wrap everything in a function then execute immediately" pattern, but I don't really like it over just executing the statements. I can't tell, but is there some reason it needs to be like this? It seems to work fine without it, but there could be corner cases I don't know about.

BryceStevenWilley (approved):
Changes LGTM and work, but I just have some nits about the new JS code.

docassemble/AssemblyLine/data/static/al_audio.js
plocket: Ah, it's possible I was just having a brain freeze. There are [some situations in which it is necessary](https://stackoverflow.com/a/19586183/14144258). For example, when you're putting an event listener on a list of elements and you can only use `var`:

```js
for ( var item of list ) { item.addEventListener('click', function(){ console.log( 'log' ) }); }
```

The above will put the event listener only on the last element. Usually these days people use `let` instead, but our aim to be compatible with older browsers doesn't allow that. Regardless, this loop is already just calling a function, so that shouldn't be a problem here.

I also left the console log in there accidentally. I can clean it all up.

plocket:
Hmm, would it help to make the text of the initial testing page very short? Letting the screen reader run to the end takes a while. I'll make a different PR with the relevant notes in comments.

docassemble/AssemblyLine/data/static/al_audio.js
BryceStevenWilley: Didn't realize we were avoid `let` because of older browsers, good to know!

BryceStevenWilley (approved):

PR #79, opened 2021-04-07
Fix #77, audio not resetting when done

### LINKED ISSUE #77: Custom audio doesn't automatically reset when audio is done playing

plocket:
Fixes #77.

Test performed:
- [ ] Start test interview
- [ ] Press 'play', icon changes to 'pause'
- [ ] Wait till audio is done
- [ ] Icon changes to 'restart'
- [ ] Tapping the icon starts the reading from the beginning

I think 'restart' is better than showing 'play' again because it gives folks some indication that this is a different situation than when they just paused the audio. It's the same reason 'stop' makes the 'restart' icon appear.


plocket:
This is independent of the previous PR.

BryceStevenWilley (approved):
LGTM, and seems like fine behavior.

PR #86, opened 2021-04-09
Cleanup some awkward language with people lists - fix #64

### LINKED ISSUE #64: If a PeopleList `xyz` has target_number=1, we should not ask for "name of the first xyz"
It should just say "name of the xyz" instead

nonprofittechy:
fix #64 

Made the PeopleList questions for children, witnesses, and generic people more responsive.

Don't ask "How many" right away...first ask if there are any.
If the list uses target_number and the target is 1, don't ask about the "first", "second", etc. Instead ask for "name of the x"

Use noun_plural with x.object_name() so that nouns are currectly pluralized/made singular in the right contexts. "Do you have any other witnesses", "name of the first witness" etc.. These will work in English and Spanish but not other languages, so we may not be able to use these questions in all interviews. Still a useful default for English and Spanish interviews.

### DIFF
--- a/docassemble/AssemblyLine/data/questions/ql_baseline.yml
+++ b/docassemble/AssemblyLine/data/questions/ql_baseline.yml
@@ -418,37 +418,71 @@ comment: |
   That will be very context-specific.
 question: |
   % if al_form_type in ['starts_case','existing_case','appeal']:
-  How many children are involved in this case?
+  Are any children involved in this case?
   % else:
-  How many children are involved in this matter?
+  Are any children involved in this matter?
   % endif  
 fields:
-  - no label: children.target_number
+  - no label: children.there_are_any
+    datatype: yesnoradio
+  - How many?: children.target_number
     datatype: integer
+    show if: children.there_are_any
+validation code: |
+  if not children.there_are_any:
+    children.target_number = 0
 ---
+# Customize this question if you know whose children they are in your interview!
+# this is just a very generic way to ask.
+id: children names without birthdate
+sets:
+  - children[0].name.first
+  - children[0].name.last
+question: |
+  % if hasattr(children, 'ask_number') and children.ask_number and children.target_number == 1:
+  Child's name
+  % else:
+  Name of first child
+  % endif
+fields:
+  - code: |
+      children[0].name_fields()
+---
+# Customize this question if you know whose children they are in your interview!
+# this is just a very generic way to ask.
 id: children names without birthdate
 sets:
   - children[i].name.first
   - children[i].name.last
 question: |
-  Name of the ${ ordinal(i) } child
+  Name of ${ ordinal(i) } child
 fields:
   - code: |
       children[i].name_fields()
 ---
 id: how many witnesses
 question: |
-  How many witnesses are there?
+  Do you have any witnesses?
 fields:
+  - no label: witnesses.there_are_any
+    datatype: yesnoradio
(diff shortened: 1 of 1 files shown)


docassemble/AssemblyLine/data/questions/ql_baseline.yml
```diff
@@ -418,54 +418,103 @@ comment: |
...
+  % if hasattr(children, 'ask_number') and children.ask_number and children.target_number == 1:
+  Child's name
+  % else:
+  Name of first child
+  % endif
+fields:
+  - code: |
+      children[0].name_fields()
 ---
+# Customize this question if you know whose children they are in your interview!
```
BryceStevenWilley, line 451: Does docassemble match the first question that will answer the field that it wants, or does it match the most specific? If it's the former, then the `children[i]` should be before the `children[0]` question.

docassemble/AssemblyLine/data/questions/ql_baseline.yml
```diff
@@ -418,54 +418,103 @@ comment: |
...
-  Name of the ${ ordinal(i) } child
+  Name of ${ ordinal(i) } child
 fields:
   - code: |
       children[i].name_fields()
 ---
 id: how many witnesses
 question: |
-  How many witnesses are there?
+  Do you have any witnesses?
```
BryceStevenWilley, line 465: How do witnesses work legally? Does a litigant 'have' a witness, or do witnesses just exist, i.e. "Do you have any witnesses?" vs "Are there any witnesses?"

docassemble/AssemblyLine/data/questions/ql_baseline.yml
```diff
@@ -418,54 +418,103 @@ comment: |
...
 yesno: x.there_is_another  
 ---
+id: name of the first person
+sets:
+  - x[0].name.first
+  - x[0].name.last
+generic object: ALPeopleList
+question: |
+  % if hasattr(x, 'ask_number') and x.ask_number and x.target_number == 1:
+  Name of ${ noun_plural(x.object_name(),1) }
```
BryceStevenWilley, line 503: Suggestion, maybe `Name of the ${ noun_plural ...`.

BryceStevenWilley (approved):
LGTM, most of my questions are just from curiosity

docassemble/AssemblyLine/data/questions/ql_baseline.yml
nonprofittechy: First it tries the most specific, and then furthest down in YAML file

docassemble/AssemblyLine/data/questions/ql_baseline.yml
nonprofittechy: On a court form, the use of "witness" is most likely to be the witness to your signature. I.e., not a witness at a trial, etc. So this is just the most common default--I can think of lots of reasons someone might want to use "witness" and have a different meaning for it, but they can customize/override for that situation.

docassemble/AssemblyLine/data/questions/ql_baseline.yml
nonprofittechy: I debated that too. Maybe @CaroRob you have an idea?

My feeling was "Name of the child", 
"Name of the witness", "Name of the Mailman" was more awkward with one person in the list than just "Name of X". But I think it's arguable which is more natural phrasing.

PR #93, opened 2021-04-16
Close #92, allow developer to set the type of name question asked

### LINKED ISSUE #92: names of opposing parties: Allow dev to set person type for each individual
Allow dev to set person type (business or person) for each individual that may be asked their name. Regarding this line in the code:

https://github.com/SuffolkLITLab/docassemble-AssemblyLine/blob/4dd5e94e1ee256527ace6b8f7c799a5e81712d99/docassemble/AssemblyLine/data/questions/ql_baseline.yml#L495

One option:
```yml
# AL Developers: To override this value, copy this block into your file and
# change 'unsure' to either 'person' or 'business'.
generic object: ALIndividual
code: |
  x.person_type = 'unsure'
---
#...
       other_parties[i].name_fields(person_or_business=other_parties[i].person_type)
```

plocket:
Closes #92.

### DIFF
--- a/docassemble/AssemblyLine/data/questions/ql_baseline.yml
+++ b/docassemble/AssemblyLine/data/questions/ql_baseline.yml
@@ -529,6 +529,12 @@ question: |
   % endif
 yesno: other_parties.there_is_another  
 ---
+# AssemblyLine: To override this value, copy this block into your file and
+# change 'unsure' to either 'person' or 'business'.
+generic object: ALIndividual
+code: |
+  x.person_type = 'unsure'
+---
 sets:
   - other_parties[i].name.first
 id: names of opposing parties
@@ -541,7 +547,7 @@ question: |
   % endif
 fields:
   - code: |
-      other_parties[i].name_fields(person_or_business='unsure')
+      other_parties[i].name_fields(person_or_business=other_parties[i].person_type)
 ---
 # TODO: consider removing default state
 # BUT: could be helpful to model how to set a default state


nonprofittechy:
This needs a placeholder value before we can consider merging it. But I'm a little confused--why is setting an invisible attribute simpler than copying the question and changing the parameter? This question will get customized a lot to be more specific.

nonprofittechy:
I'd be in support of adding a way to change the parameter in the Weaver, as that would be a bit more discoverable than copying the question.

nonprofittechy (changes requested):
Please see comments above

plocket:
I think the last comment addresses the main concern of this PR. I wasn't aware that we had meant this one to be copied by developers. My changes were made under the assumption that this block would follow the pattern of other AL question blocks. Adding it to the weaver in this case makes sense to me 👍 

As we do that, will we also remove it from AL? If not, what would be the reason for keeping it? If we do keep it, I'd love to discuss the pros and cons of the changes in this PR in the context of the block being used as it is.

Looking to the future, now that we've identified this copy/paste intention it might be worth identifying what other blocks, if any, are ones we mean developers to copy/paste as a matter of course and what to do with those as well.

nonprofittechy:
In general the goal of adding the methods like `name_fields()` is to allow copying and editing as needed without breaking translations. I think the goal is to interact with those methods by using the named keyword parameters.

We want the `question` text to mostly work out of the box, but also we want it to be easy to customize, since customization is fairly common and hard to predict in advance.

There is one more way we want to allow customization in the future which is to subclass ALIndividual, so we have a stock question for e.g., Landlord/Tenant that handles most situations involving a Tenant/Landlord, etc.

PR #96, opened 2021-04-18
Fix #95, ALPeopleList .gather() `i` not found

### LINKED ISSUE #95: ALPeopleList `foo.gather()` causes 'can't find i' error.
Code to reproduce:
```yml
include:
  - docassemble.AssemblyLine:al_package.yml
  - docassemble.MassAccess:massaccess.yml
---
objects:
  - people: ALPeopleList.using(ask_number=True, target_number=1)
---
mandatory: True
code: |
  people.gather()
---
```

Error: `There was a reference to a variable 'i' that could not be looked up in the question file`

Variable sought: `people[0].name.first`

My guess is that lines 498 and 499 need to reference `x[i]` instead of `x[0]`.
https://github.com/SuffolkLITLab/docassemble-AssemblyLine/blob/681d389d98d0455d53064eae2221025f0c1fb2ab/docassemble/AssemblyLine/data/questions/ql_baseline.yml#L496-L509

plocket:
Fixes #95. Running the test interview should show the fix working.

### DIFF
--- a/docassemble/AssemblyLine/data/questions/al_question_test.yml
+++ b/docassemble/AssemblyLine/data/questions/al_question_test.yml
@@ -8,14 +8,14 @@ modules:
   - docassemble.base.functions
 ---
 objects:
-  - people: ALPeopleList.using(target_number=1, ask_number=True)
+  - chiropractors: ALPeopleList.using(target_number=1, ask_number=True)
 ---
 mandatory: True
 code: |
   al_intro_screen
   feature_tests_1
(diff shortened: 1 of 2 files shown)


nonprofittechy:
Thanks for catching this bug! However, this fix is opposite to the code's intent. It should be [0] in all of the places instead of [i], so that the first item in the list is specially handled.

nonprofittechy (changes requested):
Please change [i] to [0]

PR #104, opened 2021-04-26
Make the "include editable copy" checkbox optional

### LINKED ISSUE #98: Feature request: Make email "editable" forms optional for PDFs
Currently, on download forms page, when requesting forms to be emailed, there is a checkbox to indicate that the user wants an editable copy of forms. For some of my programs, I don't want the user to have this option for when I am assembling PDF forms.

Can we let the developer choose whether to display the optional checkbox depending on the kind of forms we assemble?

Maybe we could do this as a parameter in the function that creates the checkbox.

nonprofittechy:
Fix #98 

Added a new keyword parameter, `show_editable_checkbox` to the `send_button_html` method. When set to "False" the "Include editable copy" checkbox will not be displayed.

### DIFF
--- a/docassemble/AssemblyLine/al_document.py
+++ b/docassemble/AssemblyLine/al_document.py
@@ -545,7 +545,14 @@ class ALDocumentBundle(DAList):
     
     return html
   
-  def send_button_html(self, key='final'):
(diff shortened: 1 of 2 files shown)


BryceStevenWilley (approved):
LGTM, haven't tested

PR #140, opened 2021-05-07
Close #138, set a default value for `has_addendum` for ALDocument

### LINKED ISSUE #138: Make `has_addendum` an optional attribute, and use `hasattr` to check its existence
Just to remove one of the required attributes at class instantiation time

plocket:
Should close #138. Not completely sure this is what was desired. Let me know if it needs to be different.


nonprofittechy:
This wasn't my original idea (I was going to wrap the references to has_addendum in hasattr wherever they are used in the class), but I think it should work just as well. Let me test to make sure.

docassemble/AssemblyLine/al_document.py
```diff
@@ -396,7 +396,7 @@ class ALDocument(DADict):
     - filename: name used for output PDF
     - title: display name for the output PDF
     - enabled
-    - has_addendum: set to False if the document never has overflow, like for a DOCX template
+    - has_addendum: The default value is False. Set to True if the document has overflow, like for a DOCX template.
```
BryceStevenWilley, line 399: The docstring now says the complete opposite as it did before. Before, it said that DOCX files shouldn't have an addendum, but now it's saying that DOCX files should have an addendum?

nonprofittechy: Good catch! Yes, DOCX files would be weird to have an addendum

plocket: Thanks @BryceStevenWilley !

PR #143, opened 2021-05-10
Typing attributes

### LINKED ISSUE #55: Add type hints for attributes in the ALDocument class
We can make type hints without assigning a value, like this:

```
class MyClass(DAObject):
  someattribute:str
  def init(self):
    pass
```

BryceStevenWilley:
Added as many types as I could to functions and classes.

Closes #55.


docassemble/AssemblyLine/al_document.py
```diff
@@ -1,5 +1,5 @@
 import re
-from typing import List, Union
+from typing import Any, Dict, List, Union
 from docassemble.base.util import log, word, DADict, DAList, DAObject, DAFile, DAFileCollection, DAFileList, defined, value, pdf_concatenate, DAOrderedDict, action_button_html, include_docx_template, user_logged_in, user_info, action_argument, send_email, docx_concatenate, reconsider, get_config, space_to_underscore, LatitudeLongitude
```
nonprofittechy, line 3: LGTM! But I think we need to import DANav here to fully take advantage of type hints

nonprofittechy (changes requested):
I think just add DANav as an import and we'll be good to merge.

docassemble/AssemblyLine/al_document.py
BryceStevenWilley: I'm not sure I understand why just including that would help, but I've added it.

nonprofittechy: At least VSCode complains about a type annotation to a type that you haven't included in the .py

Issue #65, opened 2021-03-26
Attachment download table still needs some work


nonprofittechy:
The buttons look fine vertically aligned, but without a description the placement of the form name is not quite right.

![image](https://user-images.githubusercontent.com/7645641/112568694-402ba380-8db9-11eb-9eb2-5e3dd8aa0419.png)

1. The name of the file should be top aligned and a bit larger
2. There needs to be something to fill the whitespace--could make the buttons horizontal again as one solution.

There could be other fixes to this page too. It just feels a bit awkward still.

purplesky2016:
The flow on the screen seems a bit awkward. I don't know how difficult it is to do, but something like this would be more natural:

Your form title: blah blah
`View `button
`Make Change` button
`Download `button
`Send` button
`email address` field.

Since they are all left justified, this arrangement would fix the two issues.

purplesky2016:
Here is one version
![image](https://user-images.githubusercontent.com/20231649/116456203-fa3a8300-a82f-11eb-92a5-52961ee25c8a.png)

nonprofittechy:
It is important to add a screenshot here that shows what it looks like with 3-4 forms. The "make changes" button lets you edit all of the forms, not just one.

nonprofittechy:
Decision: for now, just make the "view" button horizontally aligned with the "download" button so it looks normal if you have multiple rows in the table.

purplesky2016:
To prepare for the changes that I'm going to make, I installed  SuffolkLITLab /
docassemble-PetitionToChangeNameOfAdult on my server to see how it looks like right now, but got an error:
`JSONDecodeError: Expecting value: line 1 column 1 (char 0)`

Do I need to install something else? I installed docassemble.CARIRequest, docassemble.ALAffidavitOfIndigency and docassemble.MAPovertyScale (this one is not listed on the include section, I found it by trial and error).

Below is from the error page.

> 

> Needed definition of all_courts at 0.02885s
> Tried to run block at 0.02889s
> 
> (from docassemble.ALMassachusetts:al_massachusetts.yml)
> 
> objects:
>   - trial_court: MACourt
>   - all_courts: MACourtList.using(courts=['housing_courts','bmc','district_courts','superior_courts','land_court', 'juvenile_courts', 'p
> Log
> 
> Traceback (most recent call last):
[...]

PR #151, opened 2021-05-12
Issue #65

purplesky2016:


nonprofittechy:
image

This looks good, but can we move the "Send email" button next to the email input box instead of underneath?

purplesky2016 (commented):
Sorry for my oversight in the first place, but the top one should be removed, not the bottom one.

PR #154, opened 2021-05-14
Image shadow and interger inputbox size

purplesky2016:
As discussed, I renamed colors.css as styles.css to be used for general css rules. Two new rules are added to the bottom of the file.

This PR handles only the above changes. The other changes are for a different PR.


nonprofittechy:
This looks great--thanks @purplesky2016

PR #160, opened 2021-05-18
Close #89, add to documentation for ALDocument

### LINKED ISSUE #89: Add ALDocument example usage and attribute defaults to class docstring
[The original topic of this issue was setting defaults in the class itself. We decided `has_addendum` should have a default and we implemented that. We decided `enabled` should not have a default.

Now this issue has been converted to focus on documenting this class's behavior.]

This would be in the class itself. Ideas for defaults:
- enabled: True
- has_addendum: False

plocket:
Closes #89. I anticipate changes needed, so I figure I'll start us off. If this is a decent start and you have some specific ideas in mind for improvement, feel free to make changes.

### DIFF
--- a/docassemble/AssemblyLine/al_document.py
+++ b/docassemble/AssemblyLine/al_document.py
@@ -379,7 +379,7 @@ class DALazyAttribute(DAObject):
 
 class ALDocument(DADict):
   """
-  An opinionated collection of attachement blocks. Typically there are three:
+  An opinionated dictionary of attachment blocks. Typically there are three:
   1. The final version of a document with a signature. Use key="final" as an argument in a method.
   2. The preview version of a document with no signature. Use key="preview" as an argument in a method.
   3. An addendum of a document contained in the attribute `addendum`. E.g. `my_doc.addendum`.
@@ -443,7 +443,9 @@ class ALDocument(DADict):
   ---
   ```
   
-  Activate an addendum from the start if you already know you need one
+  For a document that may need an addendum, you must specify this when the object is created
+  or in a mandatory code block. The addendum will only be triggered if the document has "overflow"
+  in one of the fields that you specify.
   ```
   ---
   objects:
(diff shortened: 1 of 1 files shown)


docassemble/AssemblyLine/al_document.py
```diff
@@ -396,13 +396,82 @@ class ALDocument(DADict):
...
+  Activate an addendum conditionally later on if needed
+  ```
+  ---
+  objects:
+    - my_doc: ALDocument.using(filename="myDoc.pdf", title="myDoc", enabled=True)
+  ---
+  mandatory: True
+  id: interview order
+  code: |
+    # ...
```
BryceStevenWilley, line 466: For clarity, I don't think these `...` comments are needed

docassemble/AssemblyLine/al_document.py
```diff
@@ -396,13 +396,82 @@ class ALDocument(DADict):
...
+  ---
+  objects:
+    - my_doc: ALDocument.using(filename="myDoc.pdf", title="myDoc", enabled=True)
+  --- 
+  attachment:
+      variable name: my_doc[i]  # This will usually be "final" or "preview"
+      ...
+  ```
+  
+  Enable a document conditionally
```
BryceStevenWilley, line 419: Extremely minor: the addendum examples are a lot simpler than the conditionally enabled example. It might help readability if the addendum examples come before this enabling example.

BryceStevenWilley (approved):
LGTM, this is great! I have some optional ideas as comments, but honestly, we should see how this reads in the documentation site itself (with some pydoc styling) before bike-shedding it too much. Would be okay with merging as is.

docassemble/AssemblyLine/al_document.py
nonprofittechy: Note that it is actually not possible to conditionally enable the addendum. You must specify it at initialization after the recent change.

Addendum is on/off at a document level, but is only triggered if the document has overflow. I can commit an edit if that is preferred.

nonprofittechy: Clearly good evidence of why we need this documentation, so thank you @plocket

PR #155, opened 2021-05-14
Css changes for download screen

purplesky2016:

PR #162, opened 2021-05-19
Fix <div class="al_email_container">

### LINKED ISSUE #65: Attachment download table still needs some work

purplesky2016:
Close #65. Replace pull #155.


nonprofittechy:
Buttons still don't look right to me when I pull this into my playground. Any ideas?

image

nonprofittechy:
It's the right css for sure. Only thought is you mentioned al_document.css, but the CSS file is named aldocument.css? But I can see the changes you mentioned on Teams.

purplesky2016:
I am going to retract the css changes as the main goal here is to fix the py.

nonprofittechy:
Still seeing very different output on my playground after creating a new project and pulling in. Someone else want to take a look? I may be doing something wrong. @plocket ?

plocket:
For me, the table buttons are on the same line, but the 'Send' button is very tall and alignment is off at the start of the line. I can take a look to see if I can tweak a couple things.

plocket:
In order to get the alignment to work correctly, I need to know if we're going to keep the send button on the same line as the email input (created that way from UX designer feedback) or if we're going to put it on a separate line.

PR #165, opened 2021-05-19
Close #82, pause audio when buttons or links are pressed. Adds try/ca…

### LINKED ISSUE #82: Custom audio element continues playing when help button is pressed.
If you press 'play' on the page and then press 'help', the initial audio element will continue playing. The reverse is also true. It should instead... pause? Reset? We'll go with pause for now as it's the least destructive and people can restart when they want.

I'm not sure which way of doing this would be the least fragile.

Side note: In describing this issue, I'm realizing the functionality of the 'stop' button (letting the user restart) is not really signposted very well. That's a bit of a bummer. Would be great to have a designer think about that.

plocket:
…tch as that's now our practice. Refactoring seemed needed for one possible solution, but I can change most of it back since this went in another direction.

Also adds listening for the `tap` event. I thought jQuery covered taps with `click`, but I checked up on that and it doesn't seem clear that jQuery handles that.

Sorry I did so many things at once. If it's too much I can separate out the changes.

Closes #82.

### DIFF
--- a/docassemble/AssemblyLine/data/static/al_audio.js
+++ b/docassemble/AssemblyLine/data/static/al_audio.js
@@ -6,7 +6,7 @@ $(document).on('daPageLoad', function() {
     var $audio_nodes = $('.daaudio-control');
     var id_count = 1;
     for ( var audio_node of $audio_nodes ) {
-      al_js.replace_with_audio_minimal_controls( audio_node, 'page_reader_substitute_' + id_count, $audio_nodes );
+      al_js.replace_with_audio_minimal_controls( audio_node, 'page_reader_substitute_' + id_count );
       id_count++;
     }
   } catch ( error ) { console.log( 'AL audio intantiation error:', error ) }
--- a/docassemble/AssemblyLine/data/static/al_audio.js
+++ b/docassemble/AssemblyLine/data/static/al_audio.js
@@ -1,13 +1,16 @@
-$(document).on('daPageLoad', function() {
-  var $audio_nodes = $('.daaudio-control');
-  var id_count = 1;
-  for ( var audio_node of $audio_nodes ) {
-    al_js.replace_with_audio_minimal_controls( audio_node, 'page_reader_substitute_' + id_count );
-    id_count++;
-  }
-});
+// Do not overwrite already created al_js object. This may be relevant as our js gets built out.
+if ( !al_js ) { var al_js = {}; }
 
-var al_js = {};
+$(document).on('daPageLoad', function() {
+  try {
+    var $audio_nodes = $('.daaudio-control');
+    var id_count = 1;
+    for ( var audio_node of $audio_nodes ) {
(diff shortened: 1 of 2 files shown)


docassemble/AssemblyLine/data/static/al_audio.js
```diff
@@ -1,74 +1,98 @@
...
-  var id_count = 1;
-  for ( var audio_node of $audio_nodes ) {
-    al_js.replace_with_audio_minimal_controls( audio_node, 'page_reader_substitute_' + id_count );
-    id_count++;
-  }
+  try {
+    var $audio_nodes = $('.daaudio-control');
+    var id_count = 1;
+    for ( var audio_node of $audio_nodes ) {
+      al_js.replace_with_audio_minimal_controls( audio_node, 'page_reader_substitute_' + id_count, $audio_nodes );
```
BryceStevenWilley, line 9: You've added an extra function param here, but not on line 19 where you define the function itself. Is there some JS trick going on that I don't understand?

docassemble/AssemblyLine/data/static/al_audio.js
```diff
@@ -1,74 +1,98 @@
...
+  * Return functions to control audio playback for this node. */
+  var controls = {
+    play: function () {}, pause: function () {},
+    restart: function () {}, stop_and_reset: function () {},
+  };
+  
   if (!audio_node) {
     // Some screens, like the signature screen, don't have screen readers.
-    return;
+    return controls;
```
BryceStevenWilley, line 30: I'm confused about this because you're returning the controls object, but when you call the function on line 9, you're not doing anything with returned obj?

BryceStevenWilley (commented):
I guess I don't think that pausing on all clicks is useful, and would have suggested option 2 in https://github.com/SuffolkLITLab/docassemble-AssemblyLine/issues/82#issuecomment-844410217. It's just a hunch, but we don't have any user feedback on it. Pausing on help and back makes sense, because there are completely different screens, but this also stops on help text on fields and green terms too.  

Technically, if you can just not trigger on links of class `daterm`, that'd do it. But that shouldn't hold up the PR to fix the clearly broken help screen and back behavior.

docassemble/AssemblyLine/data/static/al_audio.js
plocket: Good eye. That is a remnant of another approach I left behind. Thanks for the find!

docassemble/AssemblyLine/data/static/al_audio.js
plocket: It's true, I'm not doing so right now. I did see the potential for it's use during development, though. It's a bit of premature optimization and if we want to remove it we can.

plocket:
> would have suggested option 2

Sounds like a reasonable take. Do you want to add your thoughts in the issue itself just so we can have a better paper trail, @BryceStevenWilley?

plocket:
What needs to happen next for this to move forward?

BryceStevenWilley (approved):
It's fine with me for now.

PR #172, opened 2021-05-25
Question library enhancement

### LINKED ISSUE #13: Modify general_cover_page.docx so it works with stock AssemblyLine
Right now it depends on objects from MACourts. See also: #8

### LINKED ISSUE #91: Language: Signature choice can be confusing when user is signer
From @CaroRob's comment in https://github.com/SuffolkLITLab/docassemble-209aPlaintiffMotionToModify/issues/71

> ...unless we have a question at the beginning for an advocate to help a plaintiff, it is confusing to have the option to email the form to someone else to sign.
> 
> I already said I am the plaintiff at the beginning of the interview - why would I send this form to anyone else to sign?

My reply:
> I wonder if there's some kind of wording that could be more ambiguous but still work. Maybe like 'send the form to be signed on another device'?

Kate's reply:
> I think that suggestion is good. I like having the option of emailing just in case.

My reply:
> To clarify my meaning, you might not send it to be signed by someone else, but you might send it to yourself on a different device. The current problem with that wording is that it's passive voice. I'm not sure how to improve it.

### LINKED ISSUE #111: Add a question to handle detailed user role
Some forms can be used for plaintiff/petition/appellant and the role needs to be specified on the form.

We can write a question to handle this very common need.

nonprofittechy:
Fix #91  Fix #13 Fix #111

Small additions to the question library; remove obsolete language about sending to someone else on the signature screen

### DIFF
--- a/docassemble/AssemblyLine/data/questions/al_code.yml
+++ b/docassemble/AssemblyLine/data/questions/al_code.yml
@@ -1,10 +1,18 @@
 #############################################
 # Contains code to do various actions that
 # belong in YAML and not .py files
-# 1. Email file to court
+# 1. Email file to court TODO: does this really belong here?
 # 2. Signature flow
 ##########################################################
 # Code to handle emailing court
+---
+# Placeholder
+objects:
+  - court_emails: DADict.using(auto_gather=False, gathered=True)
+---
+# Placeholder
+code: |
+  will_send_to_real_court = False
 ---
 code: |
   ready_to_email
--- a/docassemble/AssemblyLine/data/questions/al_code.yml
+++ b/docassemble/AssemblyLine/data/questions/al_code.yml
@@ -10,10 +10,6 @@
 objects:
   - court_emails: DADict.using(auto_gather=False, gathered=True)
 ---
-# Placeholder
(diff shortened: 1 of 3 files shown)


docassemble/AssemblyLine/data/questions/al_code.yml
```diff
@@ -1,11 +1,19 @@
...
 # Code to handle emailing court
 ---
+# Placeholder
+objects:
+  - court_emails: DADict.using(auto_gather=False, gathered=True)
+---
+# Placeholder
+code: |
+  will_send_to_real_court = False
+---
```
BryceStevenWilley, line 16: Can you comment more on this? It's not doing anything, since https://github.com/SuffolkLITLab/docassemble-AssemblyLine/blob/bd1ffb30c257887d27ee4addbafa806462190314/docassemble/AssemblyLine/al_general.py#L274 does the actual calculation

nonprofittechy: Aha. It looks like that function is just missing from the `__all__` list. I made the wrong fix--will correct now.

PR #175, opened 2021-05-28
65 download table

plocket:
Put send button on same line as email input, hopefully resolving some send button height issues.

### DIFF
--- a/docassemble/AssemblyLine/al_document.py
+++ b/docassemble/AssemblyLine/al_document.py
@@ -775,7 +775,6 @@ class ALDocumentBundle(DAList):
       </div>
     </div>
   '''
-    # return_str += '</div>'  # al_send_bundle
     return_str += '''
   <div class="al_email_container">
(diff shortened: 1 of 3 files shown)


purplesky2016 (commented):
In desktop view the send button is on the same line as the email address, but not in mobile.

image

plocket:
Great find, @purplesky2016 . What is the width of that mobile device?

plocket:
Also, can you double check that you've pulled in this branch. In my version, when the header text is that wide, the send button is on the same line as the input.

BryceStevenWilley:
I can't reproduce Lily's issue: the send button only goes to the next line when the screen is <= 292 pixels, which seems right, that's a very narrow screen. Works like that for me on Firefox and Chromium. `¯\_ (ツ)_/¯`

Screenshot from 2021-06-01 11-56-01

BryceStevenWilley (approved):
LGTM!

PR #183, opened 2021-06-08
Close #31, add zip row to table. Tweak and improve css.

### LINKED ISSUE #31: Allow downloading ALDocumentBundle as a .ZIP file
The "download" button for an ALDocumentBundle should have an option of downloading as a .ZIP file when there is more than one attachment in the bundle.

plocket:
Closes #31 if accepted. It's one way to do it. I abstracted a little differently, so let me know if you want to chat about it.

Now that there was longer text to see in the title cell, the buttons arrangement wasn't looking as good, so I tweaked that css a bit.

[Also ran into a bit of weirdness where using `.enabled_documents()` triggered da to look for `al_user_bundle.there_is_another` or something like that. Not sure if it's a bug or if we're just not supposed to use that method.]

### DIFF
--- a/docassemble/AssemblyLine/al_document.py
+++ b/docassemble/AssemblyLine/al_document.py
@@ -1,7 +1,7 @@
 import re
 from typing import Any, Dict, List, Union
 from docassemble.base.functions import DANav
-from docassemble.base.util import log, word, DADict, DAList, DAObject, DAFile, DAFileCollection, DAFileList, defined, value, pdf_concatenate, DAOrderedDict, action_button_html, include_docx_template, user_logged_in, user_info, action_argument, send_email, docx_concatenate, reconsider, get_config, space_to_underscore, LatitudeLongitude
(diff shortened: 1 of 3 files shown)


docassemble/AssemblyLine/al_document.py
```diff
@@ -43,25 +43,34 @@ def html_safe_str(the_string: str) -> str:
   """
   return re.sub( r'[^A-Za-z0-9]+', '_', the_string )
 
-def table_row( title, view_file:DAFile, download_file:DAFile=None, view_icon:str="eye", download_icon:str="download") -> str:
+#def table_row( title, view_file:DAFile, download_file:DAFile=None, view_icon:str="eye", download_icon:str="download") -> str:
+def table_row( title:str, buttons:list = []) -> str:
```
BryceStevenWilley, line 47:
```suggestion
def table_row( title:str, button_htmls:List[str] = []) -> str:
```
Just would make it a bit easier to understand at first glance.

nonprofittechy (changes requested):
The .zip file should contain each assembled template as a separate file, not a single PDF that combines all documents. I suggest both Word and PDF format if the original is not a PDF.

We should also make turning it on/off a parameter.

plocket:
Ready for a second review

nonprofittechy:
Looks good, thanks!

PR #188, opened 2021-06-11
Add an ALStaticDocument class to simplify using files with no fields

### LINKED ISSUE #187: Add an ALStaticDocument class that extends ALDocument
To simplify adding static files to an ALDocumentBundle

Should also assist with addressing #166

nonprofittechy:
Fix #187


docassemble/AssemblyLine/al_document.py
```diff
@@ -594,6 +594,45 @@ def overflow_value(self, field_name:str, overflow_message:str=None, preserve_new
...
+          - static_test: ALStaticDocument.using(title="Static Test", filename="static.pdf", enabled=True)
+        ---
+        objects: 
+          - bundle: ALDocumentBundle.using(elements=[static_test], filename="bundle", title="Documents to download now")
+          
+  Todo:
+      Handle files placed in /data/templates if that turns out to be useful. Likely by copying into
+      a DAFile with pdf_concatenate().
+  """
+  def init(self, *pargs, **kwargs):
```
BryceStevenWilley, line 622: I'm having issues getting the StaticFile to show up when you call `as_pdf` on a bundle. It's enabled, but it doesn't show up in the generated file, even if I use a PDF as the backing static file. Does `as_pdf` need to be overwritten too?

BryceStevenWilley (commented):
Having trouble getting the new class to work as expected, but not sure if I'm using it wrong.

docassemble/AssemblyLine/al_document.py
nonprofittechy: I think the issue was the `__str__` method, but there were some other issues with PDFs. I'm just making it an extension of DAStaticFile for now, which resolves this issue.

docassemble/AssemblyLine/al_document.py
```diff
@@ -487,6 +487,8 @@ class ALDocument(DADict):
   
   def init(self, *pargs, **kwargs):
     super(ALDocument, self).init(*pargs, **kwargs)
+    self.auto_gather=False
+    self.gathered=True
```
nonprofittechy, line 491:
```suggestion
```

BryceStevenWilley (approved):
LGTM, works for me now!

PR #202, opened 2021-06-15
Fix #192 - make sure Docassemble does not intercept the JS email input

### LINKED ISSUE #192: `index: invalid name al_doc_email_al_user_bundle` if there's a page after document bundles
**Describe the bug**
I get the system error `index: invalid name al_doc_email_al_user_bundle`
When I use the `al_question_test.yml` and add a screen after the download bundles screen

**To Reproduce**
1. On the dev server, pull the `192_name_error_demo branch` (https://github.com/SuffolkLITLab/docassemble-AssemblyLine/compare/192_name_error_demo)
2. Start the `al_question_test.yml` interview
3. Get to the downloads screen
4. Hit 'Next'
5. See the system error on the screen
6. The browser console shows two things that seem unusual, though I haven't been able to diagnose the problem with them:
   1. `Failed to execute 'setSelectionRange' on 'HTMLInputElement': The input element's type ('email') does not support selection.`
   2. `POST https://apps-dev.suffolklitlab.org/interview?i=docassemble.playground12AL189DocZip%3Aal_question_test.yml 501 (NOT IMPLEMENTED)`
6. I couldn't find anything strange in the logs

**Expected behavior**
Go on to the next page

**Screenshots**

 - Docassemble Version: 1.2.74
 - AssemblyLine Version: 2.0.19

**Additional context**
`al_doc_email_al_user_bundle` is, I believe, the string created as the `id` for the email input field. I don't see where it's being used as something like a variable/key/attribute name. I've traced it to these locations that seem relevant to its journey:
- https://github.com/SuffolkLITLab/docassemble-AssemblyLine/blob/main/docassemble/AssemblyLine/al_document.py#L798-L804
- https://github.com/SuffolkLITLab/docassemble-AssemblyLine/blob/main/docassemble/AssemblyLine/data/static/aldocument.js

nonprofittechy:


plocket (approved):
This is interesting because `al_wants_editable_input_id` does not give a `name` attribute to the checkbox even when `show_editable_checkbox` is set to `True`. It just gives it an `id`, but it still works! Maybe the `id` is a fallback after `name`.

It works great and it looks good to me. CSS is also still fine from what I can tell. Grateful there's such an easy built-in solution.

PR #204, opened 2021-06-16
Improve handling of contact methods

### LINKED ISSUE #196: The "cell" and "other" annotations should only appear if there are more than 1 phone number
https://github.com/SuffolkLITLab/docassemble-AssemblyLine/blob/c18475c131094e60959406a92c994f6be35335a4/docassemble/AssemblyLine/al_general.py#L107

### LINKED ISSUE #197: Add a "contact methods" method to an ALIndividual that uses the phones/email/other inputs and returns a string with all 3
E.g., "by telephone at ...", "by email at ...", "or "...

This is hard to neatly fit into a template, especially the appropriate use of commas/word or, especially as you cannot use a list comprehension in a DOCX template.

### LINKED ISSUE #201: Point "you need to provide at least one contact method" to the newish `field` parameter so error appears in context
https://github.com/SuffolkLITLab/docassemble-AssemblyLine/blob/4a9fa91d4a434392de92bd6e3330d16b691e9417/docassemble/AssemblyLine/data/questions/ql_baseline.yml#L465

nonprofittechy:
Fix #201 , fix #196, fix #197

### DIFF
--- a/docassemble/AssemblyLine/al_general.py
+++ b/docassemble/AssemblyLine/al_general.py
@@ -117,6 +117,24 @@ class ALIndividual(Individual):
     else:
       return ''
   
+  def contact_methods(self)-> str:
+    """Method to return a formatted string with all provided contact methods of the individual:
+        * Phone number(s)
+        * Email
+        * other method
+    Returns:
+        str: Formatted string
+    """
+    methods = []
+    if self.phone_numbers():
+      methods.append({self.phone_numbers(): word("by phone at ")})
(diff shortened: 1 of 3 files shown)


plocket (approved):
Tried several different permutations and phone numbers as well as other fields were presented as expected.

The documentation says `field` is supposed to place the message below a specific field, but I didn't see that happen. The message I saw was the message with the pink background at the top of the page floating over the page header.

Other than that, looks like it behaves as described!

Issue #36, opened 2021-02-24
Investigation: assembling documents seems slow


nonprofittechy:
Using the ALDocument class seems to slow down document assembly. Maybe something to do with it raising more exceptions than normal attachment assembly would?

Can we find a way to test if this is correct and speed it up if it really is a problem?

plocket:
I've found document display through ALDocument to be much slower than a built-in `attachment`. For a while, the table doubled the time because it was building each bundle twice - once for download and once for view - but I stopped that duplication of work. It still remained significantly slower than the built-in `attachment`.

nonprofittechy:
So that's two anecdotes anyway. It would be nice to isolate and fix it.

purplesky2016:
If we move some of the files to the Weaver's download package instead of generating them in runtime in AL, it would reduce ALDocument's workload thus speed it up. This approach could also solve a bunch of file related open issues.

nonprofittechy:
Jonathan just added a bunch of changes in DA 1.2.35 that we should experiment with. They might do a safer job of handling exceptions w/o slowing down document generation

nonprofittechy:
We might consider also generating the document in the background. We will need to trigger document generation in the interview order block that the weaver generates.

There may be other places where we can trigger definitions in interview order block to help reduce code running multiple times. That would happen in the weaver too.

plocket:
[Changelog showing 1.2.35](https://docassemble.org/docs/changelog.html)

nonprofittechy:
I think a major factor here is idempotency. If we trigger all of the "metadata" type variables of the addendum early, it could help. E.g.: trigger .enabled, .has_addendum, .overflow_fields, etc. so they are defined before the list starts getting generated. A lot of those variables won't be defined until needed.

nonprofittechy:
See #84 for probably the best overall resolution of this.

BryceStevenWilley:
I decided to dive into actual profiling results for this issue. It just seems really strange that it should take 10 seconds to assemble some relatively simple PDFs. After some setup, I now have actual profiling results on this. This mostly would drive improvements upstream in DA core, but maybe we can deduce some other workarounds
 out of this anyway.

## Results

The below graph is the call chain of functions that happen when `as_pdf()` is called. Each box is a function, the `%` is the percentage of time spent calling that function over the whole time profiled, and the `__ x` is the number of times that function is called. 

[...]

PR #205, opened 2021-06-16
Low-hanging fruit speedups

nonprofittechy:
I think this will mostly affect files with addenda. Subjectively, this already makes a different on our test file. Thanks @BryceStevenWilley for the profiling in #36


BryceStevenWilley (approved):
Looks great to me! It'd be nice to get exactly how much faster this is than baseline, but we can always go back and test it later, no need to wait.

PR #207, opened 2021-06-18
Remove default state of "MA" and make it easier to set state/country as defaults on a package-wide basis

nonprofittechy:
You could uncomment line 6 in al_question_test.yml to experiment with Canada/Canadian provinces

### DIFF
--- a/docassemble/AssemblyLine/al_general.py
+++ b/docassemble/AssemblyLine/al_general.py
@@ -49,7 +49,7 @@ class ALAddress(Address):
     ]
     if country_code:
       fields.append(
-        {"label": str(self.state_label), "field": self.attr_name("state"), "code": "states_list(country_code='{}')".format(country_code), "default": default_state}
+        {"label": str(self.state_label), "field": self.attr_name("state"), "code": "states_list(country_code='{}')".format(country_code), "default": default_state if default_state else ''}
       )
     else:
       fields.append(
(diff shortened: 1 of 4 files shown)


plocket (commented):
Unaltered defaults worked as expected, as did Canada defaults. [Labels changed correctly, as did options in the dropdown.] Putting in a non-sensical `default_state` value worked well too and simply had no state selected by default.

Unexpected behavior with manually given parameters unless I did something wrong. Saw 'State' instead of 'Province' as the `state` label.
```
users[0].address_fields(country_code="CA", default_state="ON")
```
'Postal code' was still replaced correctly.

[Demo here](https://apps-dev.suffolklitlab.org/interview?i=docassemble.playground12AL207i18n%3Aal_question_test.yml#page1)
[Branch of demo here](https://github.com/SuffolkLITLab/docassemble-AssemblyLine/tree/207_manual_address_args)

Not sure if that should block merging, so leaving this as a comment.

nonprofittechy:
Here's why that happened:

```yaml
generic object: ALAddress
template: x.state_label
# See https://docassemble.org/docs/functions.html#subdivision_type
content: |
  % if safe_subdivision_type(AL_DEFAULT_COUNTRY):
  ${ safe_subdivision_type(AL_DEFAULT_COUNTRY) }
  % else:
  State
  % endif
```

The field label is controlled by the template associated with that object, not the parameters passed to the function. That's the same as all of the other field labels.

It's hard to figure out where to add the right level of responsiveness here. I thought it would be useful to respond to the default country, but you can override the template just like you can for other labels. That way, it will not need to be customized for most real-world interviews.

nonprofittechy:
I did think of one other option, which is to have alternate templates referenced for each of the different subdivision types available. That might be the most consistent option. I have to see if there is a finite list though.

Issue #222, opened 2021-06-30
When you download an ALStaticDocument, it should be possible to override the filename


nonprofittechy:
See also #195

nonprofittechy:
Not sure if this behavior is still wrong, but it should be an easy fix?

PR #226, opened 2021-06-30
Respects the filename provided by the user in downloads

### LINKED ISSUE #195: DOCX download link should use custom filename, not `file.docx`
This might require changes in the attachment block, but if we can find a performant way to avoid that it would be better.

nonprofittechy:
Fix #195, but not yet #222

Also moved the download screen to the end to make it a little easier to find and change in the future.

### DIFF
--- a/docassemble/AssemblyLine/data/questions/al_question_test.yml
+++ b/docassemble/AssemblyLine/data/questions/al_question_test.yml
@@ -210,7 +210,7 @@ subquestion: |
   
   ---
   
-  **No zip, docx format, and hidden view buttons**
+  **No zip, docx format**
   
   ${ al_user_bundle.download_list_html( include_zip=False, format="docx") }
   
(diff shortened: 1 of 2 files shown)


plocket (approved):
Tested with a couple slightly different configurations and seems to work as expected - docx files are named. Superficially, some of the interview language may be inconsistent with current behavior and I pointed out the two spots I noticed. If the below comments don't require any changes, it looks good to me! 👍

docassemble/AssemblyLine/data/questions/al_question_test.yml
```diff
@@ -221,4 +183,39 @@ objects:
...
+  
+  ${ al_user_bundle.download_list_html(view=False) }
+  
+  ${ al_user_bundle.send_button_html(show_editable_checkbox=False) }
+    
+  ${ al_user_bundle.as_pdf() }
+  
+  ---
+  
+  **No zip, docx format, and hidden view buttons**
```
nonprofittechy, line 213:
```suggestion
  **No zip, docx format**
```

PR #229, opened 2021-07-02
Refresh `enabled` status of ALDocument objects once per page load

### LINKED ISSUE #211: `enabled` state of a document is not refreshed
If someone changes logic that makes a document enabled, the ALDocumentBundle is currently not refreshed to reflect the new state.

nonprofittechy:
Fixes #211 

This PR adds logic to store the enabled status of an ALDocument on the .cache attribute, which is reconsidered fresh on page loads (only when the attribute is needed on that page). That allows someone to dynamically change whether a document is included in the final download bundle or not by making changes on a review screen.

Side effect: you should not directly ask the user a question that sets the value of `my_doc.enabled`: instead, you can use an intermediate variable and set the value in a code block that references the intermediate variable. The code block will not be stale.

I also took the chance to do some small refactors in the al_question_test.yml file as I had to run it many times. I renamed some variables for clarity and removed some unused sections.

### DIFF
--- a/docassemble/AssemblyLine/al_document.py
+++ b/docassemble/AssemblyLine/al_document.py
@@ -403,94 +403,128 @@ class DALazyAttribute(DAObject):
           return dict()
 
 class ALDocument(DADict):
-  """
-  An opinionated dictionary of attachment blocks. Typically there are three:
-  1. The final version of a document with a signature. Use key="final" as an argument in a method.
-  2. The preview version of a document with no signature. Use key="preview" as an argument in a method.
-  3. An addendum of a document contained in the attribute `addendum`. E.g. `my_doc.addendum`.
+  """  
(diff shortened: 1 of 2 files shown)


docassemble/AssemblyLine/al_document.py
```diff
@@ -403,94 +403,128 @@ def __getstate__(self):
...
+      objects:
+        - my_doc: ALDocument.using(filename="myDoc.pdf", title="myDoc", enabled=True)
+      --- 
+      attachment:
+        variable name: my_doc[i]  # This same template will be used for the `preview` and `final` keys
+        content: |
+          Here is some content
+
+          % if i == 'final':
+          users[0].signature
```
nonprofittechy, line 469:
```suggestion
          ${ users[0].signature }
```

BryceStevenWilley (approved):
LGTM!

PR #240, opened 2021-07-15
Updated the download_html function

BryceStevenWilley:
Which hadn't been updated in the last big refactor.

Also wrapped a bunch of lines, because they were impossible to read without auto-line wrapping.

### DIFF
--- a/docassemble/AssemblyLine/al_document.py
+++ b/docassemble/AssemblyLine/al_document.py
@@ -2,7 +2,10 @@ import re
 import os
 from typing import Any, Dict, List, Union
 from docassemble.base.functions import DANav
-from docassemble.base.util import log, word, DADict, DAList, DAObject, DAFile, DAFileCollection, DAFileList, defined, value, pdf_concatenate, zip_file, DAOrderedDict, action_button_html, include_docx_template, user_logged_in, user_info, action_argument, send_email, docx_concatenate, reconsider, get_config, space_to_underscore, LatitudeLongitude, DAStaticFile
+from docassemble.base.util import log, word, DADict, DAList, DAObject, DAFile, DAFileCollection, \
+    DAFileList, defined, value, pdf_concatenate, zip_file, DAOrderedDict, action_button_html, \
+    include_docx_template, user_logged_in, user_info, action_argument, send_email, \
+    docx_concatenate, reconsider, get_config, space_to_underscore, LatitudeLongitude, DAStaticFile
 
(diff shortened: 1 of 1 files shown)


docassemble/AssemblyLine/al_document.py
```diff
@@ -931,23 +942,50 @@ def download_list_html(self, key:str='final', format:str='pdf', view:bool=True,
...
+                    download_label:str="Download", download_icon:str="download") -> str:
     """
     Returns an HTML string of a table to display all the docs
     combined into one pdf with 'view' and 'download' buttons.
     """
-    the_file = self.as_pdf(key=key)
+    if format == 'docx':
+      the_file = self.as_docx(key=key)
+    else:
+      the_file = self.as_pdf(key=key)
```
nonprofittechy, line 966: Is this the one substantive change? Trying to figure out the impact here mixed w/ line breaks.

BryceStevenWilley: Sorry, I did kinda go crazy on the line breaks.

And yes, it is substantive. Before, nothing depended on `format`, and now we're trying to preserve pdf vs docx when it's passed in.

docassemble/AssemblyLine/al_document.py
```diff
@@ -989,7 +1027,7 @@ def send_button_html(self, key:str='final', show_editable_checkbox:bool = True)-
   <div class="al_email_container">
     <span class="al_email_address '''+name+''' form-group row da-field-container da-field-container-datatype-email">
       <label for="'''+al_email_input_id+'''" class="al_doc_email col-form-label da-form-label datext-right">Email</label>
-      <input value="''' + (user_info().email if user_logged_in() else '') + '''" alt="Input box" class="form-control" type="email" size="35" name="'''+al_email_input_id+'''" id="'''+al_email_input_id+'''">
+      <input value="''' + (user_info().email if user_logged_in() else '') + '" alt="Input box" class="form-control" type="email" size="35" name="'+al_email_input_id+'" id="'+al_email_input_id+'''">
```
plocket, line 1030: If we're going to adjust how these strings are put together, is this an appropriate place to starting using the `f"..."` format?

nonprofittechy: Yes, I do think so. It would make it a lot easier to read this line. I think it can be its own PR though.

plocket (approved):
Working for me! Take my comment on string formatting as you will!

Issue #245, opened 2021-07-19
Tests: Add MVP cucumber tests interview and feature files for the document classes


plocket:
Files to create: `test_aldocument.yml`, `test_aldocument.feature`

[Also very simple template files]

nonprofittechy:
Will leave to you @plocket to decide if this issue can be closed after #251

plocket:
Thanks! There are notes in the PR itself about certain improvements that can be made to what's there, and we might make future issues about adding specific tests. Renaming and closing this for MVP.

PR #251, opened 2021-07-20
Re: #245, test ALDocument classes

plocket:
Starts by ensuring pages load with some simple method calls to classes. Adds two testing documents that require no questions to be answered. Tried to use what I've seen as the most commonly used methods. I'd like to know if this is sufficient as an MVP and, if not, what further requirements would be.

There's also some commented code that we can remove as soon as we delete the `download_html()` method.

Addresses #245


nonprofittechy (approved):
Looks great and it looks like it will be easy to add future tests. for a future PR, I think we could add more explanation for the test screens as far as what is being tested, but you can infer it from reading the code. I might also use more descriptive IDs and variable names--I prefer not to use foo/bar/baz etc. But that's beyond MVP.

Thanks! This will be a super useful addition.

Issue #238, opened 2021-07-14
Triage: Modes of Document Creation/Assembly Slowdown


BryceStevenWilley:
An extension of #36.

Most slow downs (so far) seem to fall into a few categories:

* `word_to_pdf` (made better by https://github.com/jhpyle/docassemble/releases/tag/v1.2.89)
   * though I have seen a lot more retrying libreoffice statements in logs after that release. I don't understand it enough yet though, but we need to keep an eye out. 
* `pdf_concatenate`, actually combining the pdfs together
* `defined`, and `value`, if there are a lot of variable that are checked to create the document. I have some un-PR'd fixes sitting in https://github.com/BryceStevenWilley/docassemble/tree/faster_inspect, but I need to test them a bit more before I make a PR.

This ticket is to mark specific interviews that have these slow down issues, see which are most common and when, and maybe if there are any additional types of slow downs (which you can track using the profiling tools in [ALToolbox](https://github.com/SuffolkLITLab/docassemble-ALToolbox/pull/21).

nonprofittechy:
Closed issue https://github.com/SuffolkLITLab/docassemble-AssemblyLine/issues/198 but it's worth thinking about specific solutions, like switching from pdftk to mupdf for PDF concatenation.

PR #254, opened 2021-07-21
MVP exhibit class

### LINKED ISSUE #139: Create a class to represent a list of exhibits
ALDocument as-is does not quite fit, but this class could extend ALDocument for better compatibility with ALDocumentBundle.

Features:

1. A cover page [MVP]
2. Ability to generate an index in the cover page
2. A list of DAFiles (the uploaded documents) [MVP]
3. Ability to add bates numbering
4. Ability to add separator pages between groups of DAFiles. (sometimes the DAFile will be multiple pages, or sometimes each page will be uploaded separately) QUESTION: do we want to "bundle" the files and give the bundle a title/cover page, or allow people to see all of the pages at once and insert cover pages arbitrarily?
5. Some features to reduce the size of uploaded images.

We should also add appropriate YAML block placeholders in the ql_baseline.yml file.

Some of this functionality already exists in the appeals court forms and the 209A Protection from Abuse petition.

- [x] Create simple MVP class to represent a list of exhibits
- [x] Create an MVP ALDocument wrapper for the class
- [ ] Create question library questions for uploading files
- [ ] Implement page numbering
- [ ] Implement OCRing
- [ ] Integrate method to stub + link large files for emailing
- [ ] Setup background processing events
- [ ] Investigate limiting upload size. See: https://github.com/SuffolkLITLab/docassemble-MotionToStayEviction/blob/master/docassemble/MotionToStayEviction/data/static/limit_upload_size.js (but remove some of the integration with Python/Docassemble variables--has performance implications)

nonprofittechy:
This is a simple MVP of the ALExhibitDocument class.

You can test by using the `al_question_test.yml` file, or do a minimal experiment with the signature below:

```yaml
---
objects:
  - exhibit_attachment: ALExhibitDocument.using(title="Exhibits", filename="exhibits" )
---
code: |
  exhibit_attachment.enabled = exhibit_attachment.exhibits.there_are_any
---
objects:
  - al_user_bundle: ALDocumentBundle.using(elements=[exhibit_attachment], filename="user_bundle.pdf", title="All forms to download for your records")
```

Fixes #139.

Not yet implemented: 

* OCR--it was not very reliable. But there is some code to build on in place that isn't turned on.
* Background concatenation when someone has a lot of files. Maybe we need to handle that globally though in #238 
* Running page numbers. We might not need those--sounds like those would not be necessary for any Appeals Court forms or other place with mandatory e-filing. Running page numbers will interfere with OCR which is higher priority. I will wait to implement until the OCR is stable.


plocket (commented):
Preliminary just from hand-testing: looked mostly as specified, but couldn't upload .jpg, even when I added `image/jpeg,` to the list of allowed files.

Separately, I couldn't get the provided code to work, so I must be missing something there. Kept getting 'Input not processed because question changed'. Never did find out exactly what that was about:
```yml
---
include:
  - al_package.yml
---
objects:
  - exhibit_attachment: ALExhibitDocument.using(title="Exhibits", filename="exhibits" )
---
mandatory: True
code: |
  exhibit_attachment.enabled
  end
---
code: |
  exhibit_attachment.enabled = exhibit_attachment.exhibits.there_are_any
---
objects:
  - al_user_bundle: ALDocumentBundle.using(elements=[exhibit_attachment], filename="user_bundle.pdf", title="All forms to download for your records")
---
event: end
question: the end
---
```

nonprofittechy:
I might be missing the intent of the test file, but this works:

```yaml
---
include:
  - al_package.yml
---
objects:
  - exhibit_attachment: ALExhibitDocument.using(title="Exhibits", filename="exhibits" )
---
mandatory: True
code: |
  end_screen
---
code: |
  exhibit_attachment.enabled = exhibit_attachment.exhibits.there_are_any
---
objects:
  - al_user_bundle: ALDocumentBundle.using(elements=[exhibit_attachment], filename="user_bundle", title="All forms to download for your records")
---
event: end_screen
question: the end
subquestion: |
  ${ al_user_bundle.download_list_html() }
---
```

The error you ran into was something related to idempotency, and it results in this error being logged: `index: not the same question name: ID any exhibits versus Question_166` .  I do not know how to figure out which question is Question_166--it's clearly one without an ID and it doesn't appear in the interview phrase translation file.

When I have seen errors like that in the past it was related to questions sharing an ID, but that is not the case here. `id: any exhibits` only appears once.  My only guess is that it something to do with triggering a variable that is part of the gathering process but not finishing the gathering process.

I don't fully know why it was happening with your code, but the example is not how we usually trigger a document being enabled, so it may not be necessary to track it down further.

PR #258, opened 2021-07-27
236 copy answers

nonprofittechy:
I'll create this as a draft PR for now for anyone interested. Since this is adding a lot of new variables to the namespace, I thought it might be helpful to namespace them.

There are a lot of features this does not have yet (especially around managing saved answers once you create them), but I think this is a decent MVP right now.

I suspect files will not work correctly (Jonathan says as much here: https://docassemble.org/docs/functions.html#set_session_variables) and before I make this a final PR I'm not sure if it makes sense to mark files as not private or to remove them from the answer set.


nonprofittechy:
@plocket @purplesky2016 @BryceStevenWilley if you have a chance to take a look at at this for variable names, overall style, etc., it would be appreciated. i think this is a safe merge since it's isolated code (i.e., we can always change variable names again without breaking older uses) that won't be used in production, but this is the chance to weigh in on the first draft

docassemble/AssemblyLine/data/questions/al_visual.yml
```diff
@@ -20,6 +20,10 @@ default screen parts:
     % if get_config('debug'):
     `id: ${ user_info().question_id }`  
     `Package: ${ user_info().package } ${ package_version_number }; ${ al_version }`
+    
+    [:save: Snapshot (dev)](${ url_ask(al_sessions_url_ask_snapshot) })
```
BryceStevenWilley, line 24: Do we want to add this to everyone on debug servers, or should it be hidden under a different config, so it doesn't overload debug?

docassemble/AssemblyLine/data/questions/al_saved_sessions.yml
```diff
@@ -0,0 +1,264 @@
...
+    al_sessions_save_session_snapshot_success = False
+  al_sessions_save_session_snapshot = True  
+---
+# Triggered by "Save snapshot"
+continue button field: al_sessions_snapshot_success
+question: |
+  % if al_sessions_save_session_snapshot_success:
+  Your snapshot was saved
+  % else:
+  Something went wrong saving your snapshot
```
BryceStevenWilley, line 239: The info about the failure is never saved (from the Except on like 229). Should it be logged, or just shown on this screen?

docassemble/AssemblyLine/data/questions/al_saved_sessions.yml
```diff
@@ -0,0 +1,264 @@
...
+      if not next_id:
+          break
+  al_sessions_list = sessions_list_temp
+  del sessions_list_temp
+---
+code: |
+  formatted_sessions_temp = []
+  for index, session in enumerate(al_sessions_list):
+    formatted_sessions_temp.append({
+      index: f"{session.get('title')}:{session.get('subtitle','') if session.get('subtitle','') else ''} ({ format_date(session.get('modtime')) })"
```
BryceStevenWilley, line 99: When does `session.get('subtitle', '')` return something that's false, and not just a NoneException or something similar. At a glance it's redundant to the default attribute.

docassemble/AssemblyLine/data/questions/al_saved_sessions.yml
```diff
@@ -0,0 +1,264 @@
+---
+variable name: al_sessions_url_ask_snapshot
+data:
+  - undefine: 
+    - al_sessions_save_session_snapshot
+    - al_sessions_new_session_id
+  - al_sessions_snapshot_label
+  - al_sessions_save_session_snapshot
+  - al_sessions_snapshot_success
```
BryceStevenWilley, line 9: Minor: the `al_sessions_snapshot_success` screen isn't always a success, it's more like "It worked!" or "It failed", I'd change the name to reflect that, maybe `al_sessions_snapshot_finished`?

BryceStevenWilley (approved):
LGTM, just some minor comments, would be fine with not addressing them

docassemble/AssemblyLine/data/questions/al_saved_sessions.yml
nonprofittechy: It's sometimes not present in the session information, and if it's not I just want to show the title and no `:`

Both `None` and `''` are Falsy. I think the second session.get('subtitle') does not need to have the default return though--that is potentially confusing. Good catch.

docassemble/AssemblyLine/data/questions/al_visual.yml
nonprofittechy: What are your thoughts there? I don't want to add too many config options if possible, but I'm listening. debug should definitely be turned off for production servers--both for performance reasons and security.

docassemble/AssemblyLine/data/questions/al_saved_sessions.yml
BryceStevenWilley: Ah, I think you want this instead: 
```suggestion
      index: f"{session.get('title'):session.get('subtitle') if session.get('subtitle', '') else session.get('title')} ({ format_date(session.get('modtime')) })"
```

docassemble/AssemblyLine/data/questions/al_visual.yml
BryceStevenWilley: For our use cases, it seems fine. Not sure if other people using AL 1) use debug in the dev servers (or how many have the dev/test/prod distinction), and 2) would know what this feature is: I'm reminded of the bottom feedback / About bar causing issues with ILAO recently. 

Configs are kinda confusing though. There's no a lot of discoverability. IMO either a simple "session snapshot" config (can be undocumented for now) or a minor version bump to indicate a new feature would be a fine fix.

docassemble/AssemblyLine/data/questions/al_saved_sessions.yml
nonprofittechy: no, the : needs to be outside of the `{}`. but i'll take a look and simplify it a bit.

PR #259, opened 2021-07-29
Implement custom class args. Visibly list what custom arguments are being tested on each page specifically.

plocket:
Is this clarification what you're looking for? If not, it might be useful if you could add a suggestion or just edits to either this or to the original.

Also, thanks for catching my miss on the custom arguments!


nonprofittechy (approved):
Thanks, those changes do make it much easier to follow for me. Hopefully the suggestion was helpful

PR #261, opened 2021-08-03
Close #252, alcourts tests MVP

### LINKED ISSUE #252: Tests: add test_alcourts files and MVP tests
Not sure exactly what these will look like as I've not worked with this module much before.

plocket:
Also does some transformation of the address if needed since it can come from user input. I can change that back if desired.

Having multiple pages this way ends up letting the test run faster. When we have a testing step for tapping the 'Back' button, we can get rid of some of these pages.

Close #252

### DIFF
--- a/docassemble/AssemblyLine/al_courts.py
+++ b/docassemble/AssemblyLine/al_courts.py
@@ -1,7 +1,7 @@
 #####################################
 # Package for a very simple / MVP list of courts that is mostly signature compatible w/ MACourts for now
 
-from docassemble.base.util import path_and_mimetype, Address, LatitudeLongitude, DAStaticFile, markdown_to_html, prevent_dependency_satisfaction, DAObject, DAList, DADict, log, space_to_underscore, defined
+from docassemble.base.util import path_and_mimetype, Address, LatitudeLongitude, DAStaticFile, markdown_to_html, prevent_dependency_satisfaction, DAObject, DAList, DADict, log, space_to_underscore
 from docassemble.base.legal import Court
 import pandas as pd
 import os
@@ -45,7 +45,7 @@ class ALCourt(Court):
       list.
       """
       # Avoid forcing the interview to define the court's address
-      if defined( self.attr_name('address.city') ):
+      if hasattr( self, 'address' ) and hasattr( self.address, 'city' ):
         if self.address.city in str(self.name):
           return str(self.name)
         else:
@@ -58,19 +58,7 @@ class ALCourt(Court):
       Returns a markdown formatted string with the name and address of the court.
       More concise version without description; suitable for a responsive case.
       """
-      label = f'**{ self.short_label() }**'
-      
-      # Handle inconsistent user-defined court data
-      # If address has not been left blank
-      if (( defined( self.attr_name('address.address') )
-          or defined( self.attr_name('address.city') ))
-          and self.address.on_one_line() != ', ' ):
-        # Handle no street address first line
-        street_normalized = re.sub( r'^, ', '', self.address.on_one_line() )
-        # Handle missing city name
-        address = re.sub( r', ,', ',', street_normalized )
-        label += f'[BR]{ address }'
-      return label
+      return f'**{ self.short_label() }**[BR]{ self.address.on_one_line() }'
     
     def short_description(self)->str:
       """
@@ -78,7 +66,10 @@ class ALCourt(Court):
       the description of the court, for inclusion in the results page with radio
       buttons.
       """
-      return f'{ self.short_label_and_address() }[BR]{ self.description }'
+      all_info = f'**{ self.short_label() }**'
+      if hasattr( self, 'address' ):
+        all_info = f'{ all_info }[BR]{ self.address.on_one_line() }'
+      return f'{ all_info }[BR]{ self.description }'
   
     def from_row(self, df_row, ensure_lat_long=True)->None:
       """
--- a/docassemble/AssemblyLine/al_courts.py
+++ b/docassemble/AssemblyLine/al_courts.py
@@ -1,10 +1,11 @@
 #####################################
 # Package for a very simple / MVP list of courts that is mostly signature compatible w/ MACourts for now
 
-from docassemble.base.util import path_and_mimetype, Address, LatitudeLongitude, DAStaticFile, markdown_to_html, prevent_dependency_satisfaction, DAObject, DAList, DADict, log, space_to_underscore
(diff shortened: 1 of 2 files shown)


docassemble/AssemblyLine/al_courts.py
```diff
@@ -43,25 +44,41 @@ def short_label(self)->str:
       append city name to the court name. This is good for a drop-down selection
       list.
       """
-      if self.address.city in str(self.name):
-        return str(self.name)
+      # Avoid forcing the interview to define the court's address
+      if defined( self.attr_name('address.city') ):
```
nonprofittechy, line 48: Recommend using hasattr and not `defined` wherever possible. It's possible here

docassemble/AssemblyLine/al_courts.py
```diff
@@ -43,25 +44,41 @@ def short_label(self)->str:
...
+        return str( self.name )
     
     def short_label_and_address(self)->str:
       """
       Returns a markdown formatted string with the name and address of the court.
       More concise version without description; suitable for a responsive case.
       """
-      return '**' + self.short_label() + '**' + '[BR]' + self.address.on_one_line()
+      label = f'**{ self.short_label() }**'
+
```
nonprofittechy, line 62: This behavior no longer matches the name of the method.

docassemble/AssemblyLine/data/questions/test_alcourts.yml
```diff
@@ -0,0 +1,253 @@
+---
+include:
+  - al_package.yml
+---
+metadata:
+  title: |
+    Test courts classes
+---
+# TODO: When the testing framework implements the back button,
```
nonprofittechy, line 9: Suggestion: using the back button sounds complex and might not be needed here. So designing around the current framework might be better

docassemble/AssemblyLine/al_courts.py
```diff
@@ -43,25 +44,41 @@ def short_label(self)->str:
...
+      return label
     
     def short_description(self)->str:
       """
       Returns a Markdown formatted string that includes the disambiguated name and 
       the description of the court, for inclusion in the results page with radio
       buttons.
       """
-      return '**' + self.short_label() + '**' + '[BR]' + self.address.on_one_line() + '[BR]' + self.description
+      return f'{ self.short_label_and_address() }[BR]{ self.description }'
```
nonprofittechy, line 81: Is it intentional to remove the bold formatting here?

nonprofittechy (changes requested):
I would take another look at the changes to the ALCourt class. The test looks fine.

nonprofittechy:
Something to keep in mind with this test, which I recognize isn't complete yet, is that the user input of name and address will almost always be a substitute for choosing from a list and vice versa. If the user provided the name and address of a court, they shouldn't interact with the `short_description()` etc. methods. Conversely, if they do select from a list, the address should generally be provided in the spreadsheet or database.

docassemble/AssemblyLine/al_courts.py
nonprofittechy: Never mind--I see it's inheriting from a different method now!

plocket:
> Something to keep in mind with this test, which I recognize isn't complete yet, is that the user input of name and address will almost always be a substitute for choosing from a list and vice versa. If the user provided the name and address of a court, they shouldn't interact with the `short_description()` etc. methods. Conversely, if they do select from a list, the address should generally be provided in the spreadsheet or database.

I'm not sure what this means in practice. That is, how should the code and tests be different?

Should the code and tests assume that the user will not be inputting an address? That is, that the input doesn't need to be cleaned up?

What's the relevance to the description? The tests just define the description so that there's a way to test that `short_description()` functions correctly.

Also, should I take this to mean that eventually the tests should use JSON court data? If so, whoever writes the tests and data for that needs to understand the different edge cases that need to be tested.

docassemble/AssemblyLine/al_courts.py
plocket: This was meant to display information correctly if the user omits certain pieces of the address. Is that not desired? Do we expect that addresses will only be through developers?

nonprofittechy: Let's make the method match its name. It would be confusing otherwise. Safety is a good goal but this is not an instance where it is needed.

This method would never be used in a user input situation. It is to aid with selecting the court from a list of predefined courts.

nonprofittechy:
> 
> 
> > Something to keep in mind with this test, which I recognize isn't complete yet, is that the user input of name and address will almost always be a substitute for choosing from a list and vice versa. If the user provided the name and address of a court, they shouldn't interact with the `short_description()` etc. methods. Conversely, if they do select from a list, the address should generally be provided in the spreadsheet or database.
> 
> I'm not sure what this means in practice. That is, how should the code and tests be different?
> 
> Should the code and tests assume that the user will not be inputting an address? That is, that the input doesn't need to be cleaned up?

We can happily test all of the methods. My comment was about what changes to the methods are justified.

These methods: `short_label`, `short_label_and_address`, and `short_description` are all meant to be used when the developer provides a list of courts and the user needs to choose one of the items in the list. Therefore there is no need to make any of these three methods handle user input.

That said, I think the change to make `short_label` not depend on the city is justified, because the developer might upload a list of courts without addresses. But it does not make sense to remove the address from the `short_label_and_address` method.

FYI, these were all copied blindly from the `MACourt` class, and actually they don't look like they'll ever be used in the ALCourtLoader/ALCourt class, because that class does not load all of the court data until a court is selected from the list. So maybe we should just remove the methods for now! The functionality they have should be migrated to the ALCourtLoader class.

> What's the relevance to the description? The tests just define the description so that there's a way to test that `short_description()` functions correctly.

See above--it was just one of the 3 methods that doesn't really need to handle user inputted court information.

> Also, should I take this to mean that eventually the tests should use JSON court data? If so, whoever writes the tests and data for that needs to understand the different edge cases that need to be tested.

Eventually we should test the data loading functionality. There is a sample XLSX court file in the suffolklitlab/docassemble-ALGenericJurisdiction repo.

This class is being used by one developer in Canada right now, and I think we still have a lot to learn about how it can be built out in a generic way. I don't remember if either Illinois or Louisiana are using it yet--I think they are both working with just selecting the user's county name. I'm willing to wait to see what features are needed before we add any functionality.

plocket:
Made changes. Not sure if this is what was desired. For example, if those methods are being removed, I'm not sure what this is meant to be testing at this time. There's always the option of not testing this for now and leaving testing for the future when we have more of an idea of what we want from this module.

nonprofittechy (approved):
Looks good! We can adjust tests when we make further changes to the class (no rush until we have a real-world use case), but I'm fine testing the current behavior as you do here.

PR #282, opened 2021-09-08
Resolve some idempotency issues with ALExhibit questions

### LINKED ISSUE #269: Investigate idempotency with current ALExhibitList questions
**Describe the bug**
Sometimes receive same screen twice or "Input not processed because question changed" when using the current ql_baseline questions for ALExhibitList objects.

**To Reproduce**

This can happen when running al_question_test.yml if you choose "no" to uploading documents.

**Expected behavior**
Screen continues with no toast.

**Screenshots**

image

```
some potentially relevant log info:

16:53:56 NameError exception during document assembly: name 'exhibit_attachment.exhibits[0].title' is not defined
16:54:05 Calling the as_pdf() method for bundle All forms to download for your records
16:54:05 Calling the as_pdf() method for bundle All forms to download for your records
Calling the as_pdf() method for bundle All forms to download for your records
16:54:05 Storing bundle All forms to download for your records at al_user_bundle.cache.preview
16:54:05 Calling the as_pdf() method for Word Template
16:54:05 Obtained libreoffice lock after 0.0005 seconds.
16:54:06 Finished libreoffice after 1.0945 seconds.
16:54:07 Calling the as_pdf() method for bundle All forms to download for your records
16:54:07 Storing bundle All forms to download for your records at al_user_bundle.cache.preview
16:54:07 Calling the as_pdf() method for Word Template
16:54:07 Storing main file only Word Template at word_template.cache.preview
16:54:09 index: not the same question name: ID any exhibits versus ID preview
```

 - Docassemble Version: 1.2.96
 - AssemblyLine Version: AL-2.3.0

**Additional context**

I think this might have to do with the validation code which deletes the pages.gathered attribute. Maybe using pages.reset_gathered() would be safer, and also we might need to add a `sets` and turn on `scan for variables: False` for the first "do you have any documents" screen.

nonprofittechy:
Fix #269 by adding an intermediate variable to replace direct reference to there_is_another. Currently, when someone runs al_question_test.yml, and you choose *not* to upload a file, you get the screen twice and an idempotency warning. When merged, this should let someone run the al_question_test.yml without either unexpected behavior.


BryceStevenWilley (approved):
Not tested, but LGTM!

PR #287, opened 2021-09-15
Gender enhancements

### LINKED ISSUE #278: Add an "unknown" choice for gender
https://github.com/SuffolkLITLab/docassemble-AssemblyLine/blob/2537cb1cba7ada94b2df8dcef24e5323fb2f8449/docassemble/AssemblyLine/al_general.py#L229

### LINKED ISSUE #279: Maybe we need a way to force these values to be defined for `gender_male`, `gender_female` so they work with `skip undefined`
https://github.com/SuffolkLITLab/docassemble-AssemblyLine/blob/2537cb1cba7ada94b2df8dcef24e5323fb2f8449/docassemble/AssemblyLine/data/questions/ql_baseline.yml#L747

Curious to test whether adding a method instead of a direct variable lookup would work? are methods always defined?

nonprofittechy:
Fix #279 Fix #278 

Add an "unknown" option for gender (as forms in some provinces in Canada now provide)

Make all of the gender options work better as checkboxes in PDF forms. Currently, we provide a `x.gender_male` and `x.gender_female` attribute which is defined in a code block. That method doesn't work when you assemble a PDF that has `skip undefined` turned on, unless you explicitly trigger it in your interview order block. This PR makes the `gender_male` field into a method with the `@property` decorator, which is backwards compatible with the code block but works with `skip undefined`, and only requires that `gender` is defined explicitly.

I also added matching properties for each available gender choice for inclusive PDF forms.

A small test interview below shows that the new methods work with `skip undefined`.

[what_gender.pdf](https://github.com/SuffolkLITLab/docassemble-AssemblyLine/files/7171274/what_gender.pdf)
[gender_tests.yml.txt](https://github.com/SuffolkLITLab/docassemble-AssemblyLine/files/7171281/gender_tests.yml.txt)

### DIFF
--- a/docassemble/AssemblyLine/al_general.py
+++ b/docassemble/AssemblyLine/al_general.py
@@ -274,6 +274,12 @@ class ALIndividual(Individual):
     return self.gender == 'female'
   
   @property
+  def gender_other(self):
+    """Provide True/False for 'other' gender to assist with checkbox filling
+    in PDFs with "skip undefined" turned on for forms without more inclusive options."""
+    return ((self.gender != 'male') and (self.gender != 'female'))
+  
+  @property
   def gender_nonbinary(self):
     """Provide True/False for 'nonbinary' gender to assist with checkbox filling
     in PDFs with "skip undefined" turned on."""
--- a/docassemble/AssemblyLine/al_general.py
+++ b/docassemble/AssemblyLine/al_general.py
@@ -302,7 +302,7 @@ class ALIndividual(Individual):
   def gender_self_described(self):
     """Provide True/False for 'self-described' gender to assist with checkbox filling
     in PDFs with "skip undefined" turned on."""    
-    return self.gender == 'self-described'    
+    return not (self.gender in ['prefer-not-to-say','male','female','unknown','nonbinary'])
   
   def contact_fields(self):
     """
--- a/docassemble/AssemblyLine/al_general.py
+++ b/docassemble/AssemblyLine/al_general.py
@@ -283,7 +283,6 @@ class ALIndividual(Individual):
   def gender_nonbinary(self):
     """Provide True/False for 'nonbinary' gender to assist with checkbox filling
     in PDFs with "skip undefined" turned on."""
-    
     return self.gender == 'nonbinary'
   
   @property
--- a/docassemble/AssemblyLine/al_general.py
+++ b/docassemble/AssemblyLine/al_general.py
@@ -236,6 +236,7 @@ class ALIndividual(Individual):
         {str(self.gender_nonbinary_label): 'nonbinary'},
         {str(self.gender_prefer_not_to_say_label): 'prefer-not-to-say'},
         {str(self.gender_prefer_self_described_label): 'self-described'},
+        {str(self.gender_unknown_label): 'unknown'},
     ]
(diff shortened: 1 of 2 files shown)


docassemble/AssemblyLine/al_general.py
```diff
@@ -259,7 +260,50 @@ def gender_fields(self, show_help=False):
...
+  def gender_other(self):
+    """Provide True/False for 'other' gender to assist with checkbox filling
+    in PDFs with "skip undefined" turned on for forms without more inclusive options."""
+    return ((self.gender != 'male') and (self.gender != 'female'))
+  
+  @property
+  def gender_nonbinary(self):
+    """Provide True/False for 'nonbinary' gender to assist with checkbox filling
+    in PDFs with "skip undefined" turned on."""
+
```
purplesky2016, line 286: Delete the blank line.

docassemble/AssemblyLine/al_general.py
```diff
@@ -259,7 +260,50 @@ def gender_fields(self, show_help=False):
...
+    """Provide True/False for 'female' gender to assist with checkbox filling
+    in PDFs with "skip undefined" turned on."""
+    return self.gender == 'female'
+  
+  @property
+  def gender_other(self):
+    """Provide True/False for 'other' gender to assist with checkbox filling
+    in PDFs with "skip undefined" turned on for forms without more inclusive options."""
+    return ((self.gender != 'male') and (self.gender != 'female'))
+
```
purplesky2016, line 281: Is `def gender_other(self)` still needed since line 752 is deleted?

docassemble/AssemblyLine/data/questions/ql_baseline.yml
```diff
@@ -329,6 +329,11 @@ content: |
   Self-described gender
 ---
 generic object: ALIndividual
+template: x.gender_unknown_label
+content: |
+  Unknown
+---
+generic object: ALIndividual
```
purplesky2016, line 336: I know this is not part of the changes you made, but `x.gender_female_label` seems to be defined twice.

docassemble/AssemblyLine/al_general.py
```diff
@@ -259,7 +260,50 @@ def gender_fields(self, show_help=False):
...
+  def gender_undisclosed(self):
+    """Provide True/False for 'prefer-not-to-say' gender to assist with checkbox filling
+    in PDFs with "skip undefined" turned on."""
+    return self.gender == 'prefer-not-to-say'  
+  
+  @property
+  def gender_self_described(self):
+    """Provide True/False for 'self-described' gender to assist with checkbox filling
+    in PDFs with "skip undefined" turned on."""    
+    return self.gender == 'self-described'
```
BryceStevenWilley, line 305: Unless I'm mistaken, line 242 changes the user's gender to be whatever they type in, in the self-described box. So this should probably be:
```suggestion
    return self.gender not in ['male', 'female', 'nonbinary', 'unknown', 'prefer-not-to-say']
```

docassemble/AssemblyLine/al_general.py
nonprofittechy: yes, it is for non-inclusive forms with just 3 checkboxes.

PR #309, opened 2021-10-07
Add a method to display just a user's initials.

### LINKED ISSUE #293: Add a "initials" method to ALIndividual
For use on forms that require you to initial in multiple places

```
users[0].name.first[:1] + users[0].name.middle_initial() + users[0].name.last[:1] 
```

nonprofittechy:
Fix #293 

Has an `@property` decorator so it should be called just like an attribute. E.g., `users[0].initials` without `()`

### DIFF
--- a/docassemble/AssemblyLine/al_general.py
+++ b/docassemble/AssemblyLine/al_general.py
@@ -313,6 +313,11 @@ class ALIndividual(Individual):
     Return field prompts for other contact info
     """
     pass
+  
+  @property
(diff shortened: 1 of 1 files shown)


purplesky2016:
When I tested it with the following input:
image

I got this output:
`fm.l`

Suggestions:

- Remove the dot?
- Change it to uppercase?

Question: How would the developers discover this function? Maybe add a question on the "your name" screen for them to activate it, or something like that?

PR #332, opened 2021-10-29
Limit upload size to 25 MB on browser side

### LINKED ISSUE #290: Limit upload size on ALExhibitDocument class
See: https://github.com/SuffolkLITLab/docassemble-MotionToStayEviction/blob/master/docassemble/MotionToStayEviction/data/static/limit_upload_size.js (but remove some of the integration with Python/Docassemble variables--has performance implications)

and also discussed in #139

nonprofittechy:
Fix #290 

The purpose of this fix is to prevent uploading large files on the browser frontend so that users get a better experience. Without this upload size can still be limited, but the user will not get an error before waiting for the full file to be uploaded.


purplesky2016:
I was able to trigger the error `You cannot upload files that are larger than 25 MB` with a large file.

Issue #315, opened 2021-10-20
Missing aria for the "stop" button in screen reader


nonprofittechy:
T&P 25-1

plocket:
I think this may be, in practice, the same issue as #336. Trying to find the specific description of this item.

plocket:
Just found it and it looks like the same issue.

PR #339, opened 2021-11-15
Fix #336, audio stop button aria error "button empty". Adds "value" and

### LINKED ISSUE #336: Empty button in the audio playback button
Getting an accessibility error on the WAVE AIM checker that says our audio playback button is an empty button.

Hopefully an easy fix?

plocket:
Adds "value" and "aria-label" attributes.

Closes #336. Does this also deal with #315?


plocket:
From the screenshots, I believe the steps just need more time to run. I'm increasing the timeout in our tests in a separate branch.

plocket:
I'm actually thinking that something like "Read this screen out loud", as I believe docassemble has already, may be more useful than just "play" and "stop". Maybe something to discuss together.

nonprofittechy (approved):

Issue #325, opened 2021-10-20
Add the title of the form into the top navigation bar


nonprofittechy:
See TP 58

![image](https://user-images.githubusercontent.com/7645641/138149166-3ee00106-f67d-41a6-acf5-e1639642a3fe.png)

![image](https://user-images.githubusercontent.com/7645641/138149759-6bf099d1-c023-481d-b443-f6d1c052e39f.png)

![image](https://user-images.githubusercontent.com/7645641/138149866-7f2ae3e1-a583-4c47-9093-b2a9a0888df1.png)

nonprofittechy:
@purplesky2016 would you be interested in working on re-doing the default Assembly Line (and mass access) logos to look kind of like the third picture, with the specific form name written underneath?

This is what I had so far without changing any CSS, which is not great

![image](https://user-images.githubusercontent.com/7645641/138153550-9679a4ae-3872-4fde-879f-a8b2dca76e1a.png)

PR #343, opened 2021-11-22
Fix issue #325

purplesky2016:
Changes after the call:

1. Removed line wrap for mobile navbar. 
2. The var used for the title is `interview_short_title`

**Important**: This change needs to go in tandem with removing the logo line from massaccess.yml and its css rules related to logo and title. For that matter, should the two logo files be removed as well?


nonprofittechy:
Looking good! I may not be able to merge right away before I'm back from Thanksgiving break--I want to swap out the placeholder logo with something that is less opinionated for people who haven't created their own themes yet.

PR #345, opened 2021-12-06
Fix #334

### LINKED ISSUE #334: Explicitly list all fields that can be set by `name_fields` etc in the `sets` modifier
We had an issue where the interview order block asked for users[0].address.state but the address question doesn't say it can provide that field.

nonprofittechy:
Let more idiosyncratic ways of triggering screens succeed by explicitly listing the variables that can be set on each screen in ql_baseline.yml


BryceStevenWilley (approved):
Tested, works for me. :+1: 

Not sure what to think of the fact that this will still send the user to a question if it doesn't fill in that field, i.e. the AL_DEFAULT_COUNTRY isn't US, but something is still asking for zip, so it never will get filled from these questions. Don't think it'll break anything new though.

PR #348, opened 2021-12-07
Replace CourtFormsOnline w/ AL_ORGANIZATION_TITLE

### LINKED ISSUE #346: Remove instantces of "Court Forms Online" from this package
There are 3 main places where the phrase "Court Forms Online" appears in this repo:
* the [basic questions intro screen](https://github.com/SuffolkLITLab/docassemble-AssemblyLine/blob/543814f26d59db8aada69d81f322b669fadc7184/docassemble/AssemblyLine/data/questions/ql_baseline.yml#L28-L33)
* the [sharing email/sms templates](https://github.com/SuffolkLITLab/docassemble-AssemblyLine/blob/543814f26d59db8aada69d81f322b669fadc7184/docassemble/AssemblyLine/data/questions/al_visual.yml#L239-L249)
* the [share form screen](https://github.com/SuffolkLITLab/docassemble-AssemblyLine/blob/543814f26d59db8aada69d81f322b669fadc7184/docassemble/AssemblyLine/data/questions/al_visual.yml#L163-L171), less critical since they're just default messages that the interview user can change

They're really small places, and technically can be overridden, but it's difficult to find where there are until you see them, and in general AL should be more agnostic, and these blocks can be moved as-is to [MassAccess](https://github.com/SuffolkLITLab/docassemble-MassAccess).

The open questions here that I'd need to answer before doing more with this are:
* do we all agree that we should remove "Court Forms Online" in general?
* what's a better, more generic, default to replace with "Court Forms Online", for the out-of-the-box AL interviews. I would using something like `get_config("servername")`, but it doesn't seem quite right. Maybe just "Online forms"?
* do we want to replace the phrase with something "hard coded", i.e. written in the files just like Court Forms Online is now, or do we think that a Server-wide config is a better choice (with a still hard coded fall back)?

BryceStevenWilley:
Also replaced https://courtformsonline.org with AL_ORGANIZATION_HOMEPAGE

Decided to just use `CourtFormsOnline` everywhere to be consistent. Open to trying to edit things if we think that `Court Forms Online` (with spaces) is still useful.

Note also that prod doesn't have the variable "app hompage" in the config yet, so we're still depending on that default value. 

Fixes #346.

### DIFF
--- a/docassemble/AssemblyLine/data/questions/al_visual.yml
+++ b/docassemble/AssemblyLine/data/questions/al_visual.yml
@@ -177,8 +177,7 @@ fields:
   - Message: share_interview_answers_message
     datatype: area
     default: |
-      Hi, I wanted to share my progress on a form on ,
-      "${ AL_ORGANIZATION_TITLE }".
+      Hi, I wanted to share my progress on a form on "${ AL_ORGANIZATION_TITLE }".
       If you click this link, you can follow along or finish the form for me.
     js show if: |
       val("al_how_share_link") === "email_or_sms" && val("al_sharing_type") === "share_answers"      
--- a/docassemble/AssemblyLine/data/questions/al_visual.yml
+++ b/docassemble/AssemblyLine/data/questions/al_visual.yml
@@ -4,13 +4,13 @@ objects:
 ---
 default screen parts:
   title: |
-    ${ AL_ORGANIZATION_TITLE }
+    ${ all_variables(special='metadata').get('title', AL_ORGANIZATION_TITLE) }
   short title: |
-    ${ AL_ORGANIZATION_TITLE }
+    ${ all_variables(special='metadata').get('short title', AL_ORGANIZATION_TITLE) }
   title url: |
-    ${ AL_ORGANIZATION_HOMEPAGE }
+    ${ all_variables(special='metadata').get('title url', AL_ORGANIZATION_HOMEPAGE) }
   exit url: |
-    ${ AL_ORGANIZATION_HOMEPAGE }
+    ${ all_variables(special='metadata').get('exit url', AL_ORGANIZATION_HOMEPAGE) }
   logo: |
     <div class="title-container">
       <div class="al-logo">
--- a/docassemble/AssemblyLine/data/questions/al_visual.yml
+++ b/docassemble/AssemblyLine/data/questions/al_visual.yml
@@ -1,11 +1,11 @@
 # TODO: come up with a nice-looking generic placeholder icon/label to go here
 metadata:
   title: |
-    CourtFormsOnline: MassAccess
+    ${ AL_ORGANIZATION_TITLE }: MassAccess
   short title: |
-    CourtFormsOnline
-  title url: https://courtformsonline.org  
-  exit url: https://courtformsonline.org
+    ${ AL_ORGANIZATION_TITLE }
+  title url: ${ AL_ORGANIZATION_HOMEPAGE }
+  exit url: ${ AL_ORGANIZATION_HOMEPAGE }
 ---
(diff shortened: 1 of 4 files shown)


docassemble/AssemblyLine/data/questions/al_visual.yml
```diff
@@ -170,15 +170,15 @@ fields:
...
       help you, too. Check it out at the link in this message:       
     js show if: |
       val("al_how_share_link") === "email_or_sms" && val("al_sharing_type") === "tell_friend"
   - Message: share_interview_answers_message
     datatype: area
     default: |
-      Hi, I wanted to share my progress on a form on Court Forms Online,
-      "${single_paragraph(all_variables(special='metadata').get('title','Court Forms Online'))}".
+      Hi, I wanted to share my progress on a form on ,
+      "${single_paragraph(all_variables(special='metadata').get('title', AL_ORGANIZATION_TITLE))}".
```
nonprofittechy, line 181:
```suggestion
      "${ AL_ORGANIZATION_TITLE }".
```

nonprofittechy (changes requested):
We can't use Mako in `metadata` block. I think my commit plus the suggestion should resolve but give it another try. Should be ready to merge after that.

docassemble/AssemblyLine/data/questions/al_visual.yml
```diff
@@ -170,15 +170,15 @@ fields:
...
       help you, too. Check it out at the link in this message:       
     js show if: |
       val("al_how_share_link") === "email_or_sms" && val("al_sharing_type") === "tell_friend"
   - Message: share_interview_answers_message
     datatype: area
     default: |
-      Hi, I wanted to share my progress on a form on Court Forms Online,
-      "${single_paragraph(all_variables(special='metadata').get('title','Court Forms Online'))}".
+      Hi, I wanted to share my progress on a form on ,
+      "${ AL_ORGANIZATION_TITLE }".
```
nonprofittechy, line 181:
```suggestion
      Hi, I wanted to share my progress on a form on "${ AL_ORGANIZATION_TITLE }".
```

nonprofittechy:
Going to keep working on this. `default screen parts` takes priority over `metadata`. We have a bunch of interviews that use `metadata: title` for various things. So I think we can't override `title` or `short title` with the `default screen parts` unless we also copy in the value from `metadata: title` as a starting point.

docassemble/AssemblyLine/data/questions/al_visual.yml
```diff
@@ -1,16 +1,16 @@
...
 ---
 default screen parts:
+  title: |
+    ${ AL_ORGANIZATION_TITLE }
+  short title: |
+    ${ AL_ORGANIZATION_TITLE }
+  title url: |
+    ${ AL_ORGANIZATION_HOMEPAGE }
+  exit url: |
+    ${ AL_ORGANIZATION_HOMEPAGE }
```
nonprofittechy, line 13:
```suggestion
  title: |
    ${ all_variables(special='metadata').get('title', AL_ORGANIZATION_TITLE) }
  short title: |
    ${ all_variables(special='metadata').get('short title', AL_ORGANIZATION_TITLE) }
  title url: |
    ${ all_variables(special='metadata').get('title url', AL_ORGANIZATION_HOMEPAGE) }
  exit url: |
    ${ all_variables(special='metadata').get('exit url', AL_ORGANIZATION_HOMEPAGE) }
```

nonprofittechy: This is a little convoluted, but will respect existing `metadata: title` values that are very widespread in existing interviews.

PR #352, opened 2021-12-23
Fix #347 - remove redundant intro screen

### LINKED ISSUE #347: Duplicate Basic Info / Acknowledged Inform Pages
There's a question at https://github.com/SuffolkLITLab/docassemble-AssemblyLine/blob/851cf6ae2e52ffab137e8042b5c289015e305931/docassemble/AssemblyLine/data/questions/ql_baseline.yml#L25-L78, and another at https://github.com/SuffolkLITLab/docassemble-AssemblyLine/blob/851cf6ae2e52ffab137e8042b5c289015e305931/docassemble/AssemblyLine/data/questions/ql_baseline.yml#L813-L863 that are _almost_ identical. They are different only in that:

* the higher one has an id of `basic questions intro screen`=
* the higher one uses `The CourtFormsOnline Project can help you ...` instead of `This website`
* the higher one has a `continue button field: al_intro_screen`, and the bottom doesn't
* the higher one includes "This site is not a lawyer..." disclaimer and link to LSC legal aid finder.

The former is the one trigger more often, given the continue button field, but the bottom one would be triggered if anyone ever specially tried to set `acknowledged_information_use` before the info screen. Either way, we should get rid of the redundancy, but I'd prefer to not make decisions about content.

nonprofittechy:
Note branch is wrongly titled. This fixes #347


nonprofittechy:
Going to merge--no impact change. Looks like a couple false positive test failures (timeouts) on the al_exhibit code which isn't related to this change.

PR #353, opened 2021-12-23
Address question improvements

nonprofittechy:
These two changes make the address field questions in ql_baseline just a little more robust. They [won't affect any existing repositories](https://github.com/search?p=1&q=org%3ASuffolkLITLab+mailing_address&type=Code).

No rush but might be nice to include with the next point release that has some more urgent fixes. Should be safe to merge with minimal review as it only adds new questions, doesn't replace any existing questions.

Here's a small test interview you can use:

```yaml
---
include:
  - al_package.yml
---
objects:
  - school_address: ALAddress
---
code: |
  AL_DEFAULT_COUNTRY = "GB"
---
#mandatory: True
question: |
  Mailing address
subquestion: |
  ${ users[0].mailing_address.block() }
---
#mandatory: True
question: |
  ${ defendants[0] }'s mailing address
subquestion: |
  ${ defendants[0].mailing_address }
---
mandatory: True
question: |
  ${ school_address.block() }
```

### DIFF
--- a/docassemble/AssemblyLine/data/questions/ql_baseline.yml
+++ b/docassemble/AssemblyLine/data/questions/ql_baseline.yml
@@ -466,18 +466,18 @@ fields:
     none of the above: |
       Somewhere else
     disable others: True      
-  - ${ x.address.address_label}: x.mailing_address.address
+  - ${ x.address.address_label }: x.mailing_address.address
     address autocomplete: True
-  - ${ x.address.unit_label}: x.mailing_address.unit
+  - ${ x.address.unit_label }: x.mailing_address.unit
     required: False
-  - ${ x.address.city_label}: x.mailing_address.city
-  - ${ x.address.state_label}: x.mailing_address.state
+  - ${ x.address.city_label }: x.mailing_address.city
+  - ${ x.address.state_label }: x.mailing_address.state
     code: |
       states_list(country_code=AL_DEFAULT_COUNTRY)
     default: ${ AL_DEFAULT_STATE }
-  - ${ x.address.zip_label}: x.mailing_address.zip
+  - ${ x.address.zip_label }: x.mailing_address.zip
     required: False      
-  - ${ x.address.country_label}: x.mailing_address.country
+  - ${ x.address.country_label }: x.mailing_address.country
     code: countries_list()
     default: ${ AL_DEFAULT_COUNTRY }
 ---
--- a/docassemble/AssemblyLine/data/questions/ql_baseline.yml
+++ b/docassemble/AssemblyLine/data/questions/ql_baseline.yml
@@ -706,6 +706,29 @@ fields:
   - code: |
       other_parties[i].name_fields(person_or_business='unsure')
 ---
+id: generic address
+generic object: ALAddress
+sets:
+  - x.address
+  - x.city
+  - x.zip
+  - x.unit
+  - x.state
+  - x.country
+question: |
+  What is the ${ x.object_name() }?
+fields:
+  - code: |
+      x.address_fields(country_code=AL_DEFAULT_COUNTRY, default_state=AL_DEFAULT_STATE)  
+comment: |
+  This is left as just object_name() rather than "address of object_name()" because
+  commonly the word "address" would be included in the variable name.
+
+  You should generally consider overriding this question with one that will be easier
+  to read. This is here just as a placeholder. It will be difficult to translate and
+  the variable name will not always make sense to show the user even with the help
+  of object_name().
+---
 # TODO: consider removing default state
 # BUT: could be helpful to model how to set a default state
 sets:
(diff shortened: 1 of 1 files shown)


plocket (commented):
Looking basically good. I had a couple questions and one formatting note, so I'll comment instead of accept and you can merge if it's as you intend.

`${ users[0].mailing_address.block() }` triggers `id: your address` and then triggers `id: your mailing address`, selecting 'Somewhere else' as the default answer. Similarly, `${ defendants[0] }'s mailing address` triggers `id: persons address` and then triggers `id: any mailing address`, etc. Is that the intended behavior?

Separately, `id: any mailing address` also includes two fields that look like the same value to me - 'Nation/Country' and 'Country'. The `id: persons address` doesn't offer both those options. I'm not sure which one is the intended behavior, but maybe the behaviors should match.

any_mailing_address

Formatting note:  Not a big thing, but in case you feel like cleaning it up - most of the `${ ... }` have spaces between the brackets and the contents. The labels, though, are missing the space at the end. e.g. `${ users[0].address.address_label}` instead of `${ users[0].address.address_label }`.

nonprofittechy:
> Looking basically good. I had a couple questions and one formatting note, so I'll comment instead of accept and you can merge if it's as you intend.
> 
> `${ users[0].mailing_address.block() }` triggers `id: your address` and then triggers `id: your mailing address`, selecting 'Somewhere else' as the default answer. Similarly, `${ defendants[0] }'s mailing address` triggers `id: persons address` and then triggers `id: any mailing address`, etc. Is that the intended behavior?


It is. Someone could provide their own version, but I think this is the right default behavior.
> 
> Separately, `id: any mailing address` also includes two fields that look like the same value to me - 'Nation/Country' and 'Country'. The `id: persons address` doesn't offer both those options. I'm not sure which one is the intended behavior, but maybe the behaviors should match.
> 

Nation/country is specific to UK. In the US that would be State. So that behavior is expected. The fields are not the same meaning. State is equivalent to sub locality level 2 in Google's address hierarchy and varies from country to country.

My thought was that someone can easily customize the stock .address block to add the country parameter and it would be much harder to customize the mailing_address one because it has much more complex logic. That's why the mailing_address version has all of the optional fields. In addition, mailing_address doesn't need to have jurisdiction in the primary country, whereas the address does. So giving more flexibility to mailing_address makes sense to me there too.

We can definitely see how people start to use these fields over time.

> any_mailing_address
> 
> Formatting note: Not a big thing, but in case you feel like cleaning it up - most of the `${ ... }` have spaces between the brackets and the contents. The labels, though, are missing the space at the end. e.g. `${ users[0].address.address_label}` instead of `${ users[0].address.address_label }`.

Good idea--would accept a commit suggestion on that.

docassemble/AssemblyLine/data/questions/ql_baseline.yml
```diff
@@ -425,34 +425,61 @@ fields:
...
+  - ${ users[0].address.unit_label}: users[0].mailing_address.unit
+    required: False
+  - ${ users[0].address.city_label}: users[0].mailing_address.city
+  - ${ users[0].address.state_label}: users[0].mailing_address.state
+    code: |
+      states_list(country_code=AL_DEFAULT_COUNTRY)
+    default: ${ AL_DEFAULT_STATE }
+  - ${ users[0].address.zip_label}: users[0].mailing_address.zip
+    required: False      
+  - ${ users[0].address.country_label}: users[0].mailing_address.country
```
plocket, line 451:
```suggestion
  - ${ users[0].address.address_label }: users[0].mailing_address.address
    address autocomplete: True
  - ${ users[0].address.unit_label }: users[0].mailing_address.unit
    required: False
  - ${ users[0].address.city_label }: users[0].mailing_address.city
  - ${ users[0].address.state_label }: users[0].mailing_address.state
    code: |
      states_list(country_code=AL_DEFAULT_COUNTRY)
    default: ${ AL_DEFAULT_STATE }
  - ${ users[0].address.zip_label }: users[0].mailing_address.zip
    required: False      
  - ${ users[0].address.country_label }: users[0].mailing_address.country
```

docassemble/AssemblyLine/data/questions/ql_baseline.yml
```diff
@@ -425,34 +425,61 @@ fields:
...
+  - ${ x.address.unit_label}: x.mailing_address.unit
+    required: False
+  - ${ x.address.city_label}: x.mailing_address.city
+  - ${ x.address.state_label}: x.mailing_address.state
+    code: |
+      states_list(country_code=AL_DEFAULT_COUNTRY)
+    default: ${ AL_DEFAULT_STATE }
+  - ${ x.address.zip_label}: x.mailing_address.zip
+    required: False      
+  - ${ x.address.country_label}: x.mailing_address.country
```
plocket, line 480:
```suggestion
  - ${ x.address.address_label }: x.mailing_address.address
    address autocomplete: True
  - ${ x.address.unit_label }: x.mailing_address.unit
    required: False
  - ${ x.address.city_label }: x.mailing_address.city
  - ${ x.address.state_label }: x.mailing_address.state
    code: |
      states_list(country_code=AL_DEFAULT_COUNTRY)
    default: ${ AL_DEFAULT_STATE }
  - ${ x.address.zip_label }: x.mailing_address.zip
    required: False      
  - ${ x.address.country_label }: x.mailing_address.country
```

PR #356, opened 2021-12-29
Try to prevent workflow run on tag pushes

plocket:
Based on https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#example-ignoring-branches-and-tags. May inform https://github.com/SuffolkLITLab/ALKiln/issues/380.

Not sure how we test this without just pushing to `main`.

### DIFF
--- a/.github/workflows/run_form_tests.yml
+++ b/.github/workflows/run_form_tests.yml
@@ -2,6 +2,8 @@ name: Test the interview
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


nonprofittechy (approved):

nonprofittechy:
Let's give it a try! Looks right to me. We'll wait for the next tag I think

PR #357, opened 2021-12-31
Add document method to display pdf thumbnail while avoiding layout shift

### LINKED ISSUE #311: Stop shifting the layout when displaying document previews
When showing document previews, we aren't specifying any width or height of them, so there's a few seconds before the full pdf image loads where nothing is shown on the screen, and then suddenly shifts. This kinda sucks for users, who might try clicking on "next" before the screen suddenly changes and they end up clicking on the preview. 

I think it's referred to more broadly as layout shift, and Google's [webdev page](https://web.dev/optimize-cls/) has some more good info on how to fix it.

plocket:
Closes #311. It uses an undocumented da function, `html_filter()` to get the html string we need. Not sure about the method name. You can open `test_aldocument.yml` to see it working on the first page. Just to make sure it's clear, to work, this has to hard-code the size of the thumbnail images.

I wanted to add `alt_text` as an argument too, but wasn't exactly sure how to make it work here. I needed a sensible default, but couldn't tell if it should be `None` or an empty string. We may want to think about alt text for our document displays, by the way.

I also considered having a `width_height_ratio` parameter or something, but wasn't sure how to implement that in a way that was clear. Thoughts appreciated.


plocket:
I'd like folks to take a look at #361 first and see if they like it. I like that one a bit better than this one.

PR #359, opened 2021-12-31
Return `None` when a bundle has no enabled documents.

### LINKED ISSUE #264: Handle situation with nothing to assemble better for .as_pdf() methods
Currently, the ALDocumentBundle.as_pdf() method triggers an error with pdf_concatenate when there are no enabled documents with content to assemble. This can happen if the only document is an ALExhibitDocument and the user chooses not to upload any files.

This may not be realistic to the real world, but at least the situation where the ALExhibitDocument.as_pdf() method doesn't have anything to do should be handled.

Options:
* Return None
* Assemble an "error" PDF.

plocket:
How would a developer handle this situation? Maybe we should add that to the documentation. Closes #264

### DIFF
--- a/docassemble/AssemblyLine/al_document.py
+++ b/docassemble/AssemblyLine/al_document.py
@@ -1,7 +1,7 @@
 import re
 import os
 import mimetypes
-from typing import Any, Dict, List, Union, Callable
+from typing import Any, Dict, List, Union, Callable, Optional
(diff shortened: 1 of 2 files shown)


BryceStevenWilley (approved):
LGTM. Updated the function typing signature to reflect it can be None.

PR #361, opened 2022-01-01
Take 2 of layout shift handling. Mimics and uses DAFile's .show()...

### LINKED ISSUE #311: Stop shifting the layout when displaying document previews

plocket:
[Edit: I've made a PR upstream, so it may be worth waiting for that to resolve.]

...behavior, so it may be more familiar to the developer in that way and otherwise handles ratio more directly. Most people probably won't have to mess with the nuts and bolts anyway.

Closes #311.

Still doesn't calculate the dimensions of the document itself. Haven't researched that yet and this is a decent MVP. An upstream update may or may not have to wait until we can get the dimensions of other filetypes as well.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2021 Suffolk Legal Innovation and Technology Lab
+Copyright (c) 2022 Suffolk Legal Innovation and Technology Lab
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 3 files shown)


plocket:
Is this resolved by https://github.com/jhpyle/docassemble/pull/491?

BryceStevenWilley:
IMO the upstream fix is great. I wanna do a quick check to make sure it works, which I'll get to probably today or tomorrow

PR #369, opened 2022-01-11
Test with alkiln v3 action-controlled workflow only

plocket:
This uses only the workflow file to trigger ALKiln tests. Removed package.json and other npm files and files for local development as the ALKiln will manage these now. References the ALKiln v3 branch to keep up with updates to that branch.

The tests now aren't running on push and we're working that out separately, but I am [manually triggering this action](https://github.com/SuffolkLITLab/docassemble-AssemblyLine/actions/runs/1683178312) and the tests passed.


.env.example
```diff
@@ -1,4 +0,0 @@
-# See the `bumpversion.sh` for more information on these variables
-export TEAMS_BUMP_WEBHOOK="https://example.com/webhookb2/..." 
-export TWINE_USERNAME=""
-export TWINE_PASSWORD=""
```
BryceStevenWilley, line 4: I'd prefer we hang on to this, I think I reference it in the docs for `bump_version.sh`. We should give it a more descriptive name though for sure, maybe just `.env.bumpversion.example`?

.gitignore
```diff
@@ -1,14 +0,0 @@
...
-npm-debug.log
-error*.jpg
-screenshot_*.jpg
-downloads_*
-*_lang_*.feature
-*_report_*
-_al_test_project_name.json
-.history
-__pycache__/
-docassemble.AssemblyLine.egg-info/*
```
BryceStevenWilley, line 14: Definitely no reason to delete this file. Even if some of the entries in here don't matter any more, other still do (.DS_STORE, __pycache__, etc.)

.env_example
```diff
@@ -1,23 +0,0 @@
-# Username and password to log in to a docassemble playground to run tests
```
BryceStevenWilley, line 1: I understand that this file doesn't need be a file on the filesystem anymore for the tests to run (since everything is in Github now), but isn't it still a useful piece of documentation? PLAYGROUND_EMAIL,  PLAYGROUND_PASSWORD, PLAYGROUND_ID, and EXTRA_LANGUAGES are still needed, right?

BryceStevenWilley (commented):
Changes to the workflow LGTM, but I don't get why you deleted all off the other files as well? They're definitely still useful.

.env.example
plocket: Oh, sorry, I meant to delete just the kiln env example. I'll restore that.

.gitignore
plocket: Dang, didn't realize more was added to this. I'll restore this as well and be more selective about the contents 👍

.env_example
plocket: They are needed (until v4 comes out, and then it's something else), but the setup interview creates them. I feel that if we want to support this information, can add it to the documentation. It doesn't feel like it belongs as as a separate file in every repo. It's there for local tests and, if we're keeping support for that, we'd need the package.json too, etc. Thoughts?

.gitignore
BryceStevenWilley: IMO you don't need to be selective about the contents, just leave it all there!

.env_example
BryceStevenWilley: That's fair, it is better for it to be in the documentation. +1 for keeping it removed then.

plocket:
Ready for new review, @BryceStevenWilley. How do these discussions affect the other reviews? Do you think we should keep the alkiln `.gitignore`s in those? The plan was for the setup interview to stop making those. Do you think we should add them back in instead?

BryceStevenWilley (approved):
LGTM now! Thanks

> Do you think we should keep the alkiln .gitignores in those? The plan was for the setup interview to stop making those. Do you think we should add them back in instead?

It doesn't matter too much for those other ones, AssemblyLine is the only one with modules.

PR #370, opened 2022-01-11
Match ZIP contents to download list contents

### LINKED ISSUE #190: Bundle zip should contain DOCX files if the originals are DOCX
Perhaps only if the format="docx" option is specified for the download_list_html method

nonprofittechy:
if `format="docx"` is specified, make the .zip file also have the DOCX formatted file. If `view` is set to `True`, the .zip will also contain a PDF with the DOCX.

Fix #190.

This will also allow someone to avoid Docassemble's PDF generation which can be slow when there are many files.

### DIFF
--- a/docassemble/AssemblyLine/al_document.py
+++ b/docassemble/AssemblyLine/al_document.py
@@ -917,7 +917,7 @@ class ALDocumentBundle(DAList):
   def download_list_html(self, key:str='final', format:str='pdf', view:bool=True,
       refresh:bool=True, include_zip:bool = True, view_label="View", view_icon:str="eye",
       download_label:str="Download", download_icon:str="download", zip_label:str=None,
-      zip_icon:str="file-archive", include_original_in_zip=True) -> str:
+      zip_icon:str="file-archive") -> str:
     """
     Returns string of a table to display a list
     of pdfs with 'view' and 'download' buttons.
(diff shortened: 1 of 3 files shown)


docassemble/AssemblyLine/al_document.py
```diff
@@ -908,7 +917,7 @@ def as_editable_list(self, key:str='final', refresh:bool=True) -> List[DAFile]:
   def download_list_html(self, key:str='final', format:str='pdf', view:bool=True,
       refresh:bool=True, include_zip:bool = True, view_label="View", view_icon:str="eye",
       download_label:str="Download", download_icon:str="download", zip_label:str=None,
-      zip_icon:str="file-archive") -> str:
+      zip_icon:str="file-archive", include_original_in_zip=True) -> str:
```
BryceStevenWilley, line 920: This argument isn't ever used

nonprofittechy, line 920:
```suggestion
      zip_icon:str="file-archive") -> str:
```

BryceStevenWilley:
```suggestion
      zip_icon:str="file-archive") -> str:
```

BryceStevenWilley (approved):
LGTM, you added an argument that isn't used though.

docassemble/AssemblyLine/al_document.py
nonprofittechy: Yep, I was just fixing that, sorry.

BryceStevenWilley: Nm, you caught it as the same time as I did.

PR #372, opened 2022-01-11
Add assembly_line.yml as an alias for al_package.yml

### LINKED ISSUE #38: Discussion: Revisit how we broke up different YAML files when we have had some time to get used to them
Maybe in April?

nonprofittechy:
Fix #38 

Don't think there are any other changes to the file layout that we have come up with in the last 12 months. This is an easy one that does help make the API for using AssemblyLine in your interview a little bit easier to follow.

### DIFF
--- a/docassemble/AssemblyLine/data/questions/al_package.yml
+++ b/docassemble/AssemblyLine/data/questions/al_package.yml
@@ -1,3 +1,7 @@
+################
+# Deprecated as of 1/11/2022
+# When enough packages point to assembly_line.yml, this file should turn into an alias
+# that just includes assembly_line.yml
 ---
 include:
   - al_settings.yml
--- a/docassemble/AssemblyLine/data/questions/al_package.yml
+++ b/docassemble/AssemblyLine/data/questions/al_package.yml
@@ -1,34 +1,4 @@
-################
-# Deprecated as of 1/11/2022
-# When enough packages point to assembly_line.yml, this file should turn into an alias
-# that just includes assembly_line.yml
+# Deprecated; use assembly_line.yml instead
 ---
 include:
-  - al_settings.yml
-  - al_code.yml
-  - al_saved_sessions.yml  
-  - al_visual.yml
-  - al_document.yml
-  - ql_baseline.yml
----
-modules:
-  - docassemble.ALToolbox.misc
-  - docassemble.ALToolbox.copy_button
-  - .al_general
-  - .al_document
-  - .al_courts
-  - docassemble.GithubFeedbackForm.github_issue
----  
-#############################
-# Variables used in most interviews:
-# * trial_court
-# * all_courts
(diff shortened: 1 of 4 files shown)


docassemble/AssemblyLine/data/questions/al_package_unstyled.yml
```diff
@@ -1,20 +1,4 @@
-comment: |
-  The purpose of this YAML file is to allow you to make use of all of the Assembly
-  Line standard questions, etc. with one include, but eliminate all code that 
-  changes the visual appearance of your interview.
+# DEPRECATED - use assembly_line_unstyled instead
```
BryceStevenWilley, line 1: Don't think you included this file in the branch yet

BryceStevenWilley: Everytime, you fix it the moment I comment. :laughing:

nonprofittechy: Caught it at the same moment again :)

nonprofittechy: I think I have a bad habit of opening the PR to double-check what the branch contains. I should just use compare without clicking the "open PR" button.

docassemble/AssemblyLine/data/questions/al_package.yml
```diff
@@ -1,3 +1,7 @@
+################
+# Deprecated as of 1/11/2022
+# When enough packages point to assembly_line.yml, this file should turn into an alias
+# that just includes assembly_line.yml
```
BryceStevenWilley, line 4: Any reason that you made the includes chain different between AL and AL unstyled?

* `assembly_line.yml` -> `al_package.yml` -> everything
vs
* `al_package_unstyled.yml` -> `assemble_line_unstyled.yml` -> everything

The latter seems easier, and isolates the deprecated `al_package.yml` a bit more. I don't think it breaks anything either.

nonprofittechy: My reasoning was that al_package_unstyled.yml is only referenced in a very small number of interviews, while al_package is referenced in many files. Adding the extra include adds a ms or two of load time to the interview, so I was trying to reduce impact on interviews "in the wild". But a few ms isn't a big deal. I can be persuaded to handle both the same way.

BryceStevenWilley: That makes sense. You could say that the extra ms or so is encouragement to switch to the faster loading include :wink:

nonprofittechy: Fair enough! Maybe I'll switch it to the more "correct" way now instead of waiting.

BryceStevenWilley (approved):

PR #373, opened 2022-01-11
No more debugging logging

### LINKED ISSUE #116: Remove debugging strings in ALDocument
#105 adds some caching. I'd like to leave the debug statements that go with the caching in place for a while--they only display on dev/test environment and not production, and they could help us validate the code. However, I think they could add a small slowdown, so we should remove after this code has been in use for a while.

BryceStevenWilley:
Fixes #116

IMO it's been long enough for the debugging strings. They do clog up the logs when I'm trying to quickly look for other issues.


nonprofittechy (approved):
Fair enough

PR #375, opened 2022-01-13
Reformatted with black

nonprofittechy:
`pip install black` to use on your local machine.

Black appears to be the most popular Python code formatter. See, e.g., https://deepsource.io/blog/python-code-formatters/#black. It has some very opinionated choices that wouldn't be my default, but I think it's worth just following the standard.

The only diff in this commit is that I ran `black .` in the directory containing all of the Python modules.


BryceStevenWilley:
Took a preliminary look, and the only style that might be worth a discussion is the indention; black does 4, and we decided on 2 a while back. 

But the bigger question IMO is how do we enforce it. That's the only way that a "don't think about it" formatter really works, is if we don't have to worry about running it. Usually that means a git hook that runs on commits, but that won't work when committing from docassemble. I think it would have to be a github bot that commits corrections directly to a branch (or main?) I'm not sure if that exists though.

If we can agree that black is a good formatter, then we should merge this PR so it doesn't cause a bunch of merge conflicts with other work. But it's pretty quickly going to get out of date if we can't enforce it.

docassemble/AssemblyLine/al_courts.py
```diff
@@ -2,179 +2,218 @@
...
+            "location_latitude",
+            "location_longitude",
+        }:
+            if (
+                attribute_candidate.startswith("address_")
+                and attribute_candidate.isidentifier()
+            ):
+                setattr(
+                    self.address, attribute_candidate[8:], df_row[attribute_candidate]
+                )
```
BryceStevenWilley, line 119: This is an .... interesting choice. I hope we'd be able to code around stuff like this, which is crazy unreadable

docassemble/AssemblyLine/al_general.py
```diff
@@ -1,456 +1,639 @@
...
+        )
+    else:
+        return (
+            '<i class="fa fa-'
+            + icon
+            + " fa-"
+            + size
+            + '" style="color:var(--'
+            + color
+            + ');"></i>'
```
BryceStevenWilley, line 598: some more encouragement to use f-strings I guess, haha.

BryceStevenWilley (commented):
Looks okay in general. Nothing that I can't live with. Should discuss this on a Monday meeting and then merge after.

PR #376, opened 2022-01-13
Create black-action.yml

nonprofittechy:


.github/workflows/black-action.yml
```diff
@@ -0,0 +1,25 @@
+name: black-action
+on: [push, pull_request]
```
BryceStevenWilley, line 2: What triggers `push`, any kind of push to any branch, or just a push to main? The former might get a little out of hand fast, if every push that wasn't run with black creates a PR.

BryceStevenWilley (commented):
Love it! Want to resolve when it triggers though

.github/workflows/black-action.yml
nonprofittechy: Good point. Would changing to on pull fix that?

BryceStevenWilley: I think something like [this config](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#example-using-multiple-events-with-activity-types-or-configuration) is what we want:
```suggestion
on:
  # Trigger the workflow on push or pull request,
  # but only for the main branch
  push:
    branches:
      - main
  pull_request:
    branches:
      - main
```

PR #386, opened 2022-01-31
Import `get_country()` function for Addresses

plocket:


docassemble/AssemblyLine/al_general.py
```diff
@@ -19,7 +19,7 @@
     subdivision_type,
     this_thread,
 )
-from docassemble.base.functions import DANav
+from docassemble.base.functions import DANav, get_country
```
BryceStevenWilley, line 22: Do `country_name` as well, which wasn't included.

Also, we should be changing things to only include from `base.utils`, which is apparently the stable location for DA base imports.

```suggestion
from docassemble.base.util import DANav, get_country, country_name
```

BryceStevenWilley (changes requested):

nonprofittechy (approved):

PR #395, opened 2022-02-03
Making sure that AL_ORGANIZATION_TITLE is defined

BryceStevenWilley:
We have been running into the following error on prod for a while:

```
DAErrorMissingVariable: 

    Interview has an error.  There was a reference to a variable 'AL_ORGANIZATION_TITLE' that could not be looked up in the question file (for language 'en') or in any of the files incorporated by reference into the question file.

The referer URL was http://suffolklitlab.org
```
But it's been impossible to find where it was happening from the email. I tried at one point opening and going through every single interview on court forms online, since the referer URL was suffolklitlab.org, but none of them failed on the landing page, which is what I would have expected from that referer.

Finally realized that the DA logs might have more information, and I was able to find that it's coming from this file, which some interview (still don't know which) is including. Including the `al_settings.yml` file should fix it, and have `AL_ORGANIZATION_TITLE` be defined now.

### DIFF
--- a/docassemble/AssemblyLine/data/questions/feedback.yml
+++ b/docassemble/AssemblyLine/data/questions/feedback.yml
@@ -1,6 +1,7 @@
 ---
 include:
   - docassemble.GithubFeedbackForm:feedback.yml
+  - al_settings.yml
 ---
 metadata:
   title: ${ AL_ORGANIZATION_TITLE } Feedback Form


nonprofittechy:
Good catch!

BryceStevenWilley:
Less of an issue (shouldn't crash things), but still related to this PR, I'm trying to figure out why the feedback page (when starting it from the playground) doesn't evaluate the mako, but all of the live forms on prod seem to work fine.

I didn't think mako was supported in metadata (https://github.com/SuffolkLITLab/docassemble-AssemblyLine/pull/395/files#diff-b77ee7024823f9754df884e89717b439cf31b3385365a787794f2d1afc75ec03R7), but they all look fine in the prod.

Locally:
Screenshot from 2022-02-03 11-33-09

Prod:
Screenshot from 2022-02-03 11-35-01

nonprofittechy:
You can use mako in default screen parts but not the regular metadata block.

PR #398, opened 2022-02-08
Allow setting of logo titles more flexibly. Discussion needed.

plocket:
If we want to adopt these changes, this could take care of one of our user's recently expressed needs. i.e. For the text next to the logo in the header bar, allow them to switch which item appears in which row. Currently it's locked into org name above title or short title. This PR creates variables specifically for those positions in the header.

Not sure we want this much flexibility or abstraction, but it was easy to do so I figured I'd offer it. Feel free to close if it doesn't fall within our mission. Or feel free to edit if that's more useful.

I considered making them `template`s, but I don't see a reason for them to update those mid-stream.

### DIFF
--- a/docassemble/AssemblyLine/data/questions/al_question_test.yml
+++ b/docassemble/AssemblyLine/data/questions/al_question_test.yml
@@ -14,8 +14,19 @@ code: |
 code: |
   AL_DEFAULT_STATE = "MA"
 ---
+# Allow flexible customization of header logo titles
 code: |
-  interview_short_title = "AL Question Test"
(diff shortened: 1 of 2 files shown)


nonprofittechy:
Not really a big fan of this level of customization--I think the right thing is to just customize the `logo` screen part rather than ask people to learn a new API that might only get them part of the way there. Open to discussion though.

nonprofittechy:
Closing for now. If we find that multiple people want to invert the order of the interview title and the organization name without any additional customization, this will be great to have.

PR #416, opened 2022-03-07
Dont snapshot files and filelike objects when saving a session

nonprofittechy:
Some of the instability of copying session data into a new interview comes from file permissions. This removes file objects from saved sessions to avoid those complications.

A review of this PR can just be an "it still runs" review as this is still a developer-oriented feature. I'm going to continue to make additional small PRs to refine different aspects of the feature and eventually make it usable by members of the public.

### DIFF
--- a/docassemble/AssemblyLine/data/questions/al_saved_sessions.yml
+++ b/docassemble/AssemblyLine/data/questions/al_saved_sessions.yml
@@ -159,6 +159,13 @@ code: |
     'simple_filtered_vars_tmp',
     'al_sessions_url_ask_snapshot',
     'al_sessions_url_ask_fast_forward',
+    'AL_ORGANIZATION_HOMEPAGE',
+    'AL_ORGANIZATION_TITLE',
+    'Any',
+    'Dict',
+    'List',
+    'Optional',
+    'al_logo',
     # Maybe we will want these to be optional? Depends on how good review screen is.
     # 'docket_number',
     # 'docket_numbers',
@@ -167,6 +174,12 @@ code: |
     #'user_started_case',
   ]
 ---
+code: |
+  def file_like(obj):
+    return isinstance(obj, DAFile) or isinstance(obj, DAFileCollection) or isinstance(obj, DAFileList) or isinstance(obj, ALDocument) or isinstance(obj, ALDocumentBundle)
+---
+need:
+  - file_like
 # For save snapshot
 code: |
   filtered_vars_tmp = all_variables(simplify=False)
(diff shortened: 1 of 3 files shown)


docassemble/AssemblyLine/data/questions/al_saved_sessions.yml
```diff
@@ -167,31 +174,32 @@ code: |
...
-  # TODO: is it worth making this recursive and traversing all attributes?
-  for var in filtered_vars_tmp:
-    if isinstance(var, DAFile) or isinstance(var, DAFileCollection) or isinstance(var, DAFileList):
-      try:
-        var.set_attributes(private=False)
-      except:
-        pass
+    simple_filtered_vars_tmp.pop(var, None)   
   
   al_simple_filtered_vars = simple_filtered_vars_tmp
```
purplesky2016, line 192: 1. Do we need `simple_filtered_vars_tmp`? If we use `al_simple_filtered_vars` to begin with, it will save a few lines.
2. I meant to ask this question on Monday's call but we didn't get to it. Most packages are unique, why do we include saved sessions from outside the current interview? Is it intentional? 

1. The current session is listed on the source session. Should it be removed?

1.  Remove the down arrow as shown in the image since it's redundant and doesn't do anything?

image

purplesky2016 (commented):
When I ran `manage_sessions.yml`, the "To a new session of" field listed a few options that I don't recognize except one. I picked that one but got an error, which is true, that package is not in my playground. But why did it make to the list? Is it because it's in another playground on the same server?

`DANotFoundError: Interview docassemble.LRFGuideMe:data/questions/HousingGuide.yml not found`

nonprofittechy:
> When I ran `manage_sessions.yml`, the "To a new session of" field listed a few options that I don't recognize except one. I picked that one but got an error, which is true, that package is not in my playground. But why did it make to the list? Is it because it's in another playground on the same server?
> 
> `DANotFoundError: Interview docassemble.LRFGuideMe:data/questions/HousingGuide.yml not found`

It's using the list of "advertised" / dispatch / shortcut interviews on the server to build that list (see https://docassemble.org/docs/config.html#dispatch). It's in the `config.yml` file for the server. There's not any other good way to see what interviews are installed on the server.

docassemble/AssemblyLine/data/questions/al_saved_sessions.yml
nonprofittechy: > 1. Do we need `simple_filtered_vars_tmp`? If we use `al_simple_filtered_vars` to begin with, it will save a few lines.

Yes we do, it's to avoid Docassemble stopping running the code block too early. 

>     2. I meant to ask this question on Monday's call but we didn't get to it. Most packages are unique, why do we include saved sessions from outside the current interview? Is it intentional?

Yes. This feature is eventually designed to let you copy answers from one interview into another interview. It is true that it will only work for fields that have the same name.

>     3. The current session is listed on the source session. Should it be removed?

Yes. I'll make changes like that in a new PR, though. I want to make it a lot simpler overall, actually.

>     4. Remove the down arrow as shown in the image since it's redundant and doesn't do anything?
> 
> 
> image

That's a docassemble CSS bug for the combobox input that I've also noticed recently.

purplesky2016:
Just a thought: would it be helpful if we allow the user to choose if they want to use this functionality for current interview only, or for all all interviews on the server? I suspect the first case may be no less popular than the second; in that case it would save the trouble of having to scan thru unrelated sessions by the user.

nonprofittechy:
> Just a thought: would it be helpful if we allow the user to choose if they want to use this functionality for current interview only, or for all all interviews on the server? I suspect the first case may be no less popular than the second; in that case it would save the trouble of having to scan thru unrelated sessions by the user.

That's a good idea. Also, see my notes here: https://github.com/SuffolkLITLab/docassemble-AssemblyLine/issues/421

PR #428, opened 2022-03-29
Add additional safety to as_editable_list() method of ALDocumentBundle

### LINKED ISSUE #426: Error emailing a document that has an ALExhibitList attachment - gap in `as_editable_list()` method?
User feedback:

> I currently have the following written for email sending attachments= [al_user_bundle.as](http://al_user_bundle.as/)_editable_list() and noticed that the exhibit upload doesn’t work without getting this error:
> Interview has an error. There was a reference to a variable ‘exhibit_attachment_two.pdf’ that could not be looked up in the question file (for language ‘en’) or in any of the files incorporated by reference into the question file.
> Is there a way to set it up that the uploaded file isn’t converted into a PDF? I noticed that it works when I use as_pdf_list() but I would like the template to remain as a docx

It sounds like `as_editable_list()` is just directly referencing the .pdf attribute and we should add a little more safety there

nonprofittechy:
Fix #426

Don't assume that if a document isn't a DOCX there is a PDF version of it--instead, check and return the closest thing to an "editable" document that we can for the given file.

Also added a semantically equivalent as_docx_list() method that mirrors as_pdf_list and will call the generally safe as_docx() method for each file in the bundle instead of directly referencing the .docx or .pdf attribute. Adding this as a separate method in case someone was relying on the behavior of as_editable_list.

### DIFF
--- a/docassemble/AssemblyLine/test_al_document.py
+++ b/docassemble/AssemblyLine/test_al_document.py
@@ -1,0 +1,17 @@
+import unittest
+from docassemble.base.util import DAFile
+from .al_document import ALDocument, ALDocumentBundle
+
+class test_dont_assume_pdf(unittest.TestCase):
+  def test_upload_pdf(self):
+    doc1 = ALDocument("doc1", title="PDF 1", filename="pdf_doc_1", enabled=True, has_addendum=False)
+    doc1["final"] = DAFile("test_aldocument_pdf_1.pdf")
+    doc2 = ALDocument("doc2", title="DOCX 2", filename="docx_doc_1", enabled=True, has_addendum=False)
+    doc2["final"] = DAFile("test_aldocument_docx_1.docx")
+    al_doc_bundle = ALDocumentBundle("al_doc_bundle", elements=[doc1, doc2], title="Multiple docs", filename="multi_bundle_1", enabled=True)
+    new_list = al_doc_bundle.as_editable_list()
+    self.assertEqual(len(new_list), 2)
+    pass
+
+if __name__ == '__main__':
+  unittest.main()
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -1,0 +1,21 @@
+name: Run python only unit tests
+
+on:
+  push:
+  workflow_dispatch:
+
+jobs:
+  test-assemblyline:
+    runs-on: ubuntu-latest
+    name: Run python only unit tests
+    env:
+      ISUNITTEST: true
+    steps:
+      - run: sudo apt-get update && sudo apt-get -y install libcurl4-openssl-dev build-essential python3-dev libldap2-dev libsasl2-dev slapd ldap-utils tox lcov valgrind libzbar0
+      - name: Check
+        uses: actions/checkout@v2
+      - run: pip install virtualenv
+      - run: virtualenv -p python3.8 venv
+      - run: venv/bin/pip install -r docassemble/ALWeaver/requirements.txt
+      - run: venv/bin/pip install --editable .
+      - run: venv/bin/python3 -m unittest discover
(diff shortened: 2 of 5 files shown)


nonprofittechy:
Unit tests sound good and I think no rush

BryceStevenWilley:
Didn't want to spin too long on this, but I did add a unit test that, when run on `main`, gave the following error:
```
E
======================================================================
ERROR: test_upload_pdf (AssemblyLine.test_al_document.test_dont_assume_pdf)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/brycew/Developer/LITLab/docassemble-AssemblyLine/docassemble/AssemblyLine/test_al_document.py", line 12, in test_upload_pdf
    new_list = al_doc_bundle.as_editable_list()
  File "/home/brycew/Developer/LITLab/docassemble-AssemblyLine/docassemble/AssemblyLine/al_document.py", line 1044, in as_editable_list
    editable.append(doc.docx if hasattr(doc, "docx") else doc.pdf)
  File "/home/brycew/Developer/LITLab/litlab_venv/lib/python3.8/site-packages/docassemble/base/core.py", line 547, in __getattr__
    raise docassemble.base.functions.this_thread.misc['pending_error']
docassemble.base.error.DAAttributeError: name 'test_aldocument_pdf_1.pdf.pdf' is not defined

----------------------------------------------------------------------
Ran 1 test in 0.092s

FAILED (errors=1)
```
and has no errors on this branch. So we're good to merge here.

docassemble/AssemblyLine/test_al_document.py
```diff
@@ -0,0 +1,37 @@
+import unittest
+from docassemble.base.util import DAFile
+from .al_document import ALDocument, ALDocumentBundle
+
+
+class test_dont_assume_pdf(unittest.TestCase):
+    def test_upload_pdf(self):
```
nonprofittechy, line 7: I'm not sure I understand what this is testing.

The original error that prompted this PR was that I assumed:

1. The contents of the bundle were generated by an attachment block that produced a [DAFileCollection](https://docassemble.org/docs/objects.html#DAFileCollection) object
2. If the .docx attribute didn't exist for the DAFileCollection object, the .pdf attribute would exist

But this assumption failed if it wasn't a DAFileCollection, or if the filetype wasn't convertable to PDF and wasn't originally a DOCX.

But this could still be fine as-is and we can open a new issue to review and add unit tests

nonprofittechy:
> Didn't want to spin too long on this, but I did add a unit test that, when run on `main`, gave the following error:
> 
> ```
> E
> ======================================================================
> ERROR: test_upload_pdf (AssemblyLine.test_al_document.test_dont_assume_pdf)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/home/brycew/Developer/LITLab/docassemble-AssemblyLine/docassemble/AssemblyLine/test_al_document.py", line 12, in test_upload_pdf
>     new_list = al_doc_bundle.as_editable_list()
>   File "/home/brycew/Developer/LITLab/docassemble-AssemblyLine/docassemble/AssemblyLine/al_document.py", line 1044, in as_editable_list
>     editable.append(doc.docx if hasattr(doc, "docx") else doc.pdf)
>   File "/home/brycew/Developer/LITLab/litlab_venv/lib/python3.8/site-packages/docassemble/base/core.py", line 547, in __getattr__
>     raise docassemble.base.functions.this_thread.misc['pending_error']
> docassemble.base.error.DAAttributeError: name 'test_aldocument_pdf_1.pdf.pdf' is not defined
> 
> ----------------------------------------------------------------------
> Ran 1 test in 0.092s
> 
> FAILED (errors=1)
> ```
> 
> and has no errors on this branch. So we're good to merge here.

Sounds good--this is worth us merging, because it does confirm that the as_editable_list() will work with a regular DAFile instead of a DAFileCollection. But I don't know if the test is exactly right yet.

docassemble/AssemblyLine/test_al_document.py
BryceStevenWilley: Gotcha, I never realized that attachment blocks were DAFileCollections. I do think this PR handles "if it wasn't a DAFileCollection", but I see what you mean, probably should have tested with a jpg or some other file type.

PR #432, opened 2022-04-04
Add ALDocumentUpload class to allow adding a simple uploaded file to an ALDocumentBundle

### LINKED ISSUE #427: Create a simpler class / set of questions for ALExhibitDocument
ALExhibitDocument has features like:

- cover page
- auto-numbering
- individual document titles
- multiple pages and multiple documents

That don't always make sense. Many interviews have a single file that they want the user to upload, like a court notice, where the title is known in advance.

We just want a simple wrapper around an upload to use there that lets you drop the uploaded file into an `ALDocumentBundle`. Maybe call the class `DAUploadedFile`?

nonprofittechy:
Fix #427 

Limitations:
- The ALDocumentUpload can only contain one DAFile

### DIFF
--- a/docassemble/AssemblyLine/al_document.py
+++ b/docassemble/AssemblyLine/al_document.py
@@ -43,6 +43,7 @@ __all__ = [
     "ALTableDocument",
     "ALUntransformedDocument",
     "unpack_dafilelist",
+    "ALDocumentUpload",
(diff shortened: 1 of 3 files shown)


BryceStevenWilley (approved):
LGTM!

PR #434, opened 2022-04-05
Add a default gender of a business

### LINKED ISSUE #423: Fix the pronoun_possessive, etc. functions when the person is a business
We can do this with a small YAML shim to handle x.gender when it's a business

nonprofittechy:
Fix #423 

This will fix methods like `pronoun_possessive` if the person selected "business" on the screen that asks "Is this a business, or a person?"

```yaml
---
include:
  - assembly_line.yml
---
mandatory: True
code: |
  other_parties.gather()
---
mandatory: True
question: |
  ${ other_parties[0].pronoun_possessive("apple") }
```

### DIFF
--- a/docassemble/AssemblyLine/data/questions/ql_baseline.yml
+++ b/docassemble/AssemblyLine/data/questions/ql_baseline.yml
@@ -1670,3 +1670,8 @@ generic object: ALDocumentBundle
 code: |
   x.there_is_another = False
 ---
+generic object: ALIndividual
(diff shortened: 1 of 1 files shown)


purplesky2016 (approved):

purplesky2016:
Is there a test yml to test this change?

purplesky2016:
I was able to trigger this question and display `x.gender`' s value `other`. Good.

PR #440, opened 2022-04-06
Remove the unused ALDocumentBundleDict class

### LINKED ISSUE #404: Remove `ALDocumentBundleDict` class
https://github.com/SuffolkLITLab/docassemble-AssemblyLine/blob/6428f2b88d144443f0fe11438841b806ea32e194/docassemble/AssemblyLine/al_document.py#L1314

This class has never been used. It is probably more confusing than helpful to have it, and it's a pretty trivial addition if we decide we want it.

nonprofittechy:
Fix #404. 

This class was designed to meet a need that hasn't surfaced yet. It isn't referenced anywhere in [any public github repository](https://github.com/search?q=ALDocumentBundleDict&type=code) other than our documentation.

Instead of the ALDocumentBundleDict class, someone can just add multiple ALDocumentBundle variables or create their own list or dictionary to display them to the user--the class-specific features are redundant because the ALDocumentBundle can already be nested.

It also has never been tested. Removing it so people don't accidentally use it.

### DIFF
--- a/docassemble/AssemblyLine/al_document.py
+++ b/docassemble/AssemblyLine/al_document.py
@@ -32,7 +32,6 @@ __all__ = [
     "ALAddendumFieldDict",
     "ALDocumentBundle",
     "ALDocument",
-    "ALDocumentBundleDict",
     "ALStaticDocument",
     "safeattr",
     "label",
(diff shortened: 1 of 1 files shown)


plocket (approved):
Looks great! Do we need to re-generate the documentation or was that manual?

nonprofittechy:
> Looks great! Do we need to re-generate the documentation or was that manual?

We would need to regenerate or manually edit the documentation. I'll work on that next.

PR #441, opened 2022-04-06
Add a language_fields() method to ALIndividual

### LINKED ISSUE #272: Add a standard language picker method, like `language_fields`
I think it should be possible to customize with a list of the ~ 7 most common languages for any given organization + an "other" input.

Here's an example I've used for Mass Advocates for Children:

image

nonprofittechy:
Fix #272 

Test interview:

```yaml
---
include:
  - assembly_line.yml
---
mandatory: True
question: |
  You speak ${ users[0].language_name() }
```


plocket (approved):
Looks good to me 👍

PR #446, opened 2022-04-06
More convenient place to customize `auto_labeler` in the ALExhibitDocument class

### LINKED ISSUE #394: Add an `auto_labeler` parameter you can access when initializing an `ALExhibitDocument` object
It's a bit awkward to customize this right now. Would be useful to let you provide the labeler lambda in the `objects` block.

nonprofittechy:
A little esoteric, but when you use the ALExhibitDocument class, it defaults to giving each exhibit a cover page like `A`, `B`, etc. The `auto_labeler` attribute lets you override this function to use a different function that takes a list index and returns a suitable sequential label, like [item_label](https://docassemble.org/docs/functions.html#item_label).

This PR just makes it easier to specify the `auto_labeler` attribute when you create an ALExhibitDocument object. Specifically, you can now do it in the `objects` block.

Fix #394

Also spotted a bug in the _update_labels() method.

### DIFF
--- a/docassemble/AssemblyLine/al_document.py
+++ b/docassemble/AssemblyLine/al_document.py
@@ -1448,7 +1448,7 @@ class ALExhibitList(DAList):
             auto_labeler = self.auto_labeler
 
         for index, exhibit in enumerate(self.elements):
-            exhibit.label = self.auto_labeler(index)
+            exhibit.label = auto_labeler(index)
 
     def ocr_ready(self) -> bool:
         """
@@ -1513,6 +1513,7 @@ class ALExhibitDocument(ALDocument):
         include_exhibit_cover_pages (bool): flag to control whether cover pages are included with each separate exhibit
         add_page_numbers (bool): Flag that controls whether the as_pdf() method
           also will add Bates-stamp style page numbers and labels on each page.
+        auto_labeler (callable): a Lambda or Python function that will be used to label exhibits.          
 
     Todo:
         * Method of making a safe link in place of the attachment (e.g., filesize limits on email)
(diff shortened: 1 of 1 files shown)


docassemble/AssemblyLine/al_document.py
```diff
@@ -1448,7 +1448,7 @@ def _update_labels(self, auto_labeler: Callable = None) -> None:
             auto_labeler = self.auto_labeler
 
         for index, exhibit in enumerate(self.elements):
-            exhibit.label = self.auto_labeler(index)
+            exhibit.label = auto_labeler(index)
```
BryceStevenWilley, line 1451: I don't think this change is needed? You don't pass anything to the `auto_labeler` parameter on [link 1495](https://github.com/SuffolkLITLab/docassemble-AssemblyLine/pull/446/files#diff-db23667d014599821c0250a04d5ffbbcf327b89266e017c6f6304c30a19ff8efR1495), which means this will error our trying to call the function of `None`.

From the rest of the PR, it looks like you should just remove that parameter, and only use `self.auto_labeler`?

nonprofittechy: Maybe YAGNI applies here. My thought was to give people an option to manually call `_update_labels()` in some circumstance I haven't foreseen.

I'm not sure where it would error out though? On line 1448+1449, I was previously checking for a `None` value of `auto_labeler` and if it was null, `auto_labeler` is set to `self.auto_labeler`.

BryceStevenWilley: Ah, gotcha, my eyes skipped over the parts of the function not in this PR. That's fine, IMO. Testing now.

PR #452, opened 2022-04-22
Making our theme accessible

BryceStevenWilley:
Changes made:
* added a simple alt text to our logo image. There's a lot of differing information out there about what alt text to use (I found [this flow chart](https://www.w3.org/WAI/tutorials/images/decision-tree/) when developing), but I think going by the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/alt#icons_or_logos), and [this good blog post](https://davidmacd.com/blog/what-is-pure-decoration-alt-text-in-wcag.html) by a WCAG member, is a good idea. Those say to use some simple alt text (just the project's name) on images and logos that aren't purely decorative
* Make the audio buttons have the same names as aria labels. 
From https://www.w3.org/WAI/WCAG21/Techniques/failures/F96:

    > "When speech input users interact with a web page, they usually speak
    a command followed by the reference to some visible label (like the
    text in a button or a link, or the text labelling some input). If the
    visible label does not match the accessible name of the control, speech
    users may be unable to directly activate that control."

    The options to fix this are to:
    * remove "Listen" (bad for non-assistive tech users IMO), 
    * make all of the aria-labels "Listen" (bad for assistive tech users),
    * or just make the visible text match the aria-labels.

* Since our footer is a slightly darker white, the default blue link text doesn't quite have enough contrast (was at 4.2 something, and needed to be 4.5), so I changed the link color in the footer just slightly.

### DIFF
--- a/docassemble/AssemblyLine/data/static/al_audio.js
+++ b/docassemble/AssemblyLine/data/static/al_audio.js
@@ -98,7 +98,7 @@ al_js.replace_with_audio_minimal_controls = function( audio_node, id ) {
 
 // The DOM structure for every AL audio element with custom controls
 var audio_contents_html = '\
-  <button class="media-action play btn btn-sm btn-outline-secondary" aria-label="play" value="play">\
+  <button class="media-action play btn btn-sm btn-outline-secondary" aria-label="Listen" value="play">\
     <i class="fas fa-volume-up"></i><span>&nbsp;Play&nbsp;</span>\
     <i class="fas fa-play"></i>\
   </button>\
--- a/docassemble/AssemblyLine/data/static/al_audio.js
+++ b/docassemble/AssemblyLine/data/static/al_audio.js
@@ -99,15 +99,15 @@ al_js.replace_with_audio_minimal_controls = function( audio_node, id ) {
 // The DOM structure for every AL audio element with custom controls
 var audio_contents_html = '\
   <button class="media-action play btn btn-sm btn-outline-secondary" aria-label="play" value="play">\
-    <i class="fas fa-volume-up"></i><span>&nbsp;Listen&nbsp;</span>\
+    <i class="fas fa-volume-up"></i><span>&nbsp;Play&nbsp;</span>\
     <i class="fas fa-play"></i>\
   </button>\
   <button class="media-action restart btn btn-sm btn-outline-secondary" aria-label="restart" value="restart">\
-    <i class="fas fa-volume-up"></i><span>&nbsp;Listen&nbsp;</span>\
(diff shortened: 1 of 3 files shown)


purplesky2016:
Looking good!

docassemble/AssemblyLine/data/static/al_audio.js
```diff
@@ -99,15 +99,15 @@ al_js.replace_with_audio_minimal_controls = function( audio_node, id ) {
 // The DOM structure for every AL audio element with custom controls
 var audio_contents_html = '\
   <button class="media-action play btn btn-sm btn-outline-secondary" aria-label="play" value="play">\
-    <i class="fas fa-volume-up"></i><span>&nbsp;Listen&nbsp;</span>\
+    <i class="fas fa-volume-up"></i><span>&nbsp;Play&nbsp;</span>\
```
nonprofittechy, line 102:
```suggestion
    <i class="fas fa-volume-up"></i><span>&nbsp;Listen&nbsp;</span>\
```

docassemble/AssemblyLine/data/static/al_audio.js
```diff
@@ -99,15 +99,15 @@ al_js.replace_with_audio_minimal_controls = function( audio_node, id ) {
 // The DOM structure for every AL audio element with custom controls
 var audio_contents_html = '\
   <button class="media-action play btn btn-sm btn-outline-secondary" aria-label="play" value="play">\
```
nonprofittechy, line 101:
```suggestion
  <button class="media-action play btn btn-sm btn-outline-secondary" aria-label="Listen" value="play">\
```

BryceStevenWilley:
There was some conversation about header size today. The main issue is that the main question specifically has the `h3` class applied to it, which makes the h1 header look like an h3 header.

```html
<div class="da-page-header">
  <h1 class="h3" id="daMainQuestion">What is your name?</h1>
  <div class="daclear"></div>
</div>
<div class="da-subquestion">
  <h1>header 1</h1>
  <h2>header 2</h2>
  <h3>header 3</h3>
  <h4>header 4</h4>
</div>
```
On the default docassemble, it looks like this:
Screenshot from 2022-04-25 19-24-41

The best approach to this is the somewhat hacky thing of making all headers within the `da-subquestion` div look like one header below, i.e. make `h2` have the `.h3` class, etc.

BryceStevenWilley:
> Except I don't see much color difference for the footer links. Maybe add !important in the rule

That is intentional, it's a very slight change. I checked in the browser tools, and it's using the right color value from this file.

nonprofittechy (approved):

PR #470, opened 2022-05-04
Mypy action

BryceStevenWilley:
Actually starts running mypy on each commit though the existing unittest action. Necessary to make sure that our types are correct.

Started off in the [suggested state](https://breadcrumbscollector.tech/mypy-how-to-use-it-in-my-project/) of just ignoring errors that occur from third party libraries. Eventually we might want to add stubs for other libraries like the country stuff, or we can make it so that those libraries have types themselves (:eyes: docassemble :eyes:). Probably worth it at some point to look through everything and make sure there are types where there should be.


nonprofittechy:
This looks great--how can I best test, and what's our current status as far as type safety? Will the current code pass/fail?

nonprofittechy (approved):

Issue #53, opened 2021-03-15
Start thinking about how to represent the form's data in a reusable way--not just as a PDF


nonprofittechy:
We eventually want to be able to store the information captured in the interview in a structured way, that captures all of the same information that goes onto the PDF. We may want to do this two different ways: a version with all of the data, and an anonymized version with just a summary for running reports.

Some ideas: extend the ALDocument class so that it also assembles a "shadow" JSON file. We could both send this to the court and store in database in searchable JSONB format. This would let us capture the information at the same time we generate the final document for the user/court.

We may also want to format and structure this data to match the fields in ECFv4 or v5; align with the profileMe hackathon experiment; look to NCSC and Schema.org.

At this point: we just need to start thinking about the problem and try to design our classes to allow room for this expansion.

nonprofittechy:
@BryceStevenWilley one of our ideas was that we should try sending this data to the court even before the court has a place for it to go. Is there an existing metadata field where arbitrary JSON or XML data could be attached?

plocket:
I think one of the strengths of ALDocument is that it has one job and does it well - assembling the documents. It might be better to keep this data building functionality separate from ALDocument, even if ALDocument ends up utilizing the data that's created.

nonprofittechy:
> I think one of the strengths of ALDocument is that it has one job and does it well - assembling the documents. It might be better to keep this data building functionality separate from ALDocument, even if ALDocument ends up utilizing the data that's created.

Valid point. Worth keeping simplicity and clarity of purpose in mind as a design goal. I might need to clarify though--what we are thinking about here is the same data that goes into the PDF, but in a structured format rather than a picture that a person has to read. It should not contain different data. I think this could be part of ALDocument in the same way that the addendum code is part of ALDocument. These representations tend to be machine-readable documents, although keeping them indexable and searchable in a database row would be nice. XML is the old wave; JSON is the newer approach but not dominant. For some examples of people who have created structured representations of legal data:

* http://www.akomantoso.org/
* https://www.ncsc.org/services-and-experts/areas-of-expertise/court-statistics/national-open-court-data-standards-nods
* https://www.oasis-open.org/committees/tc_home.php?wg_abbrev=legaldocml
* https://schema.org/
* https://github.com/jhpyle/docassemble-profileme

[...]

PR #476, opened 2022-05-09
Add export and import of interview data in JSON format

### LINKED ISSUE #291: Create a one-click simplified vars download button in developer HUD
Something like what's visible in Snapshot (dev)

nonprofittechy:
This adds an "Import answer set" feature which allows loading a JSON file into the current interview session.

image

And a corresponding "Export as json" button on the answer set page.

image

It relies on the same code that allows the Docassemble API endpoint to set interview answers, and may be considered a bit risky. We need to make sure that our "deny list" approach is safe enough for this purpose.

While the "export" function is turned on by default, the "import" feature has been placed behind a configuration flag.

To turn it on on a given server, add this to your configuration:

```yaml
assembly line:
  enable answer set imports: True
```

Fix #291 

This also relates to #53 -


purplesky2016:
1. The `Export `button opens a tab, rather than download a file.
2. Where can I access the `upload a json file` screen? I thought it would be on the `load answer set` screen, but it's not there.
3. I saw quite a few issues related to json file export/import, so apparently it is a desired functionality. I am only wondering if this is the idea location for it. Do we expect the `save/load answer set` users to deal with a json file, or is it for a different group of users? Can we discuss this on Monday?

nonprofittechy:
> * The `Export `button opens a tab, rather than downloads a file.

I can keep looking into that. It's a good point. This actually depends on the user's browser settings, but I think we can force it to be a download.

> * Where can I access the `upload a json file` screen? I thought it would be on the `load answer set` screen, but it's not there.

Because of the limited way that the action system works in Docassemble, I made the "import" action its own menu item:

image

This menu item will only appear if the configuration entry 

```yaml
assembly line:
  enable answer set imports: True
 ```
exists on the server. This is turned on on the apps-dev server but if you're testing on MLRI server it won't be turned on. LMK if you need me to configure that on the server you're using for testing.

> * I saw quite a few issues related to json file export/import, so apparently it is a desired functionality. I am only wondering if this is the idea location for it. Do we expect the `save/load answer set` users to deal with a json file, or is it for a different group of users? Can we discuss this on Monday?

Yes, let's chat on Monday. There's no real rush here. I'm actually already using the functionality in some other projects though, so I hope we're able to merge it.

I think there are a couple of relevant uses cases here. 

1. The main one that is fairly abstract is just data transparency--letting users observe the data that we collect and "take out" their data so they retain control and ownership over it.
2. Eventually-data reuse without requiring accounts. This isn't that yet, since the UI only surfaces this for logged in users. But the functions could be used that way.
3. Keeping client data in the same location as other client information. This is mostly for attorneys and advocates who already use case management systems. Allowing export and import lets them keep the client information attached to the case management system in a reusable form, instead of just a PDF which would require re-typing for a new form and then a linked server which the attorney doesn't control.
4. Re-use across servers, where the existing answer set falls down.
5. Backup and persistence of data.

I can imagine some further afield use cases, like using this to store data in the court's docket (would require some mapping of field names). But I think the use cases that are described above really benefit from an interface that lets the user import and export information. I think none of these use cases are going to be used by a majority of users, but they might be used quite a lot by power users like attorneys, community organizers and advocates.

purplesky2016:
Yes, please enable it on the MLH dev server whenever you have a chance, no rush.

I added this topic to Monday's agenda.

nonprofittechy:
> Yes, please enable it on the MLH dev server whenever you have a chance, no rush.

Added

nonprofittechy:
@purplesky2016 will defer to you but I think we identified a number of useful next steps for this code today, but nothing blocking merging--is that right?

nonprofittechy:
Some of our follow-up suggestions are in:

#491 #490 #489

PR #494, opened 2022-05-17
Create a feature flag for the Answer Set feature

### LINKED ISSUE #488: Put "answer set" feature behind an opt-in configuration flag

nonprofittechy:
Fix #488, make the new answer set feature opt-in

### DIFF
--- a/docassemble/AssemblyLine/data/questions/al_visual.yml
+++ b/docassemble/AssemblyLine/data/questions/al_visual.yml
@@ -79,11 +79,17 @@ code: |
     {"url": url_ask(['al_start_over_confirmation','al_start_over']),
     "label": "Start over"
     },
+  ]
+  
+  # Answer set feature is behind a server-wide configuration option
+  if get_config('assembly line',{}).get('enable answer sets'):
+    al_menu_items_tmp.append([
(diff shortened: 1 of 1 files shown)


docassemble/AssemblyLine/data/questions/al_visual.yml
```diff
@@ -79,18 +79,25 @@ code: |
     {"url": url_ask(['al_start_over_confirmation','al_start_over']),
     "label": "Start over"
     },
+  ]
```
BryceStevenWilley, line 82: This works, but only on brand new interviews. I imagine if someone decided to turn it on they'd want it to apply to all existing interviews.

I never know the best solution for problems like this though. I guess `reconsider: al_menu_items`?

BryceStevenWilley: Yeah. Making the list is fine, tbh. Most expensive thing is probably just accessing the config probably.

nonprofittechy: I'll put this on the block that computes the value of `al_menu_items` instead, because someone might not want to use our menu items at all.

BryceStevenWilley (approved):

PR #501, opened 2022-05-25
Add property getter for phone_numbers

BryceStevenWilley:
Allows for use to get the full list of phone numbers in things like `showifdef`, which
doesn't like functions.
Used the `all_phone_numbers` so it doesn't cause any breaking changes to `phone_numbers`.

Currently causing some issues in prod, since the Interpreternotice includes a `showifdef("users[0].phone_numbers()")` in the DOCX template, and there's no simple way to fix that error without something like this.

### DIFF
--- a/docassemble/AssemblyLine/al_general.py
+++ b/docassemble/AssemblyLine/al_general.py
@@ -397,6 +397,11 @@ class ALIndividual(Individual):
         else:
             return ""
 
+    @property
+    def all_phone_numbers(self) -> str:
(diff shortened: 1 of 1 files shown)


BryceStevenWilley:
That's fair, it had been working with some quick tests, but it could very well blow up in the future. I hadn't realized that `phone_numbers()` returns an empty string by default, just tested it and that's definitely the better solution. Shouldn't need any fix in AL core then.

PR #508, opened 2022-06-08
Exhibit improvements

BryceStevenWilley:
Needed to finish up the ALExhibitDocument, adding bates numbering and OCR. I want to test a bit more on the dev and test servers, but running on a local docassemble, things were fairly stable. The default for `auto_ocr` and `add_page_numbers` is false, so this won't change any existing behavior or make anything unstable yet.

Here's the test interview I used:
```yaml
include:
  - al_package.yml
---
mandatory: True
code: |
  exhibits.exhibits.gather()
  
  if not exhibits.exhibits.ocr_ready():
    waiting_screen
  else:
    final_screen
---
event: waiting_screen
question: Waiting
reload: True
---
event: final_screen
question: |
  Here is your document.
subquestion: |
  ${ single_pdf_bundle_1.as_pdf(pdfa=True) }
---
objects:
  # No addenda, all start enabled
  - single_pdf_bundle_1: ALDocumentBundle.using(elements=[exhibits], pdfa=True, title="One pdf", filename="single_pdf_1" )
---
objects:
  - exhibits: ALExhibitDocument.using( title="PDF 1", filename="pdf_doc_1", enabled=True, auto_ocr=True, add_page_numbers=True)
```

### DIFF
--- a/docassemble/AssemblyLine/al_document.py
+++ b/docassemble/AssemblyLine/al_document.py
@@ -1352,7 +1352,7 @@ class ALDocumentBundle(DAList):
 class ALExhibit(DAObject):
     """Class to represent a single exhibit, with cover page, which may contain multiple documents representing pages.
     Atributes:
-        elements (list): List of individual DAFiles representing uploaded images or documents.
+        pages (list): List of individual DAFiles representing uploaded images or documents.
         cover_page (DAFile | DAFileCollection): (optional) A DAFile or DAFileCollection object created by an `attachment:` block
           Will typically say something like "Exhibit 1"
         label (str): A label, like "A" or "1" for this exhibit in the cover page and table of contents
@@ -1366,12 +1366,36 @@ class ALExhibit(DAObject):
         if not hasattr(self, "starting_page"):
             self.start_page = 1
 
-    def add_numbers(self, starting_page=None) -> None:
+    def _start_ocr(self):
         """
-        Todo:
-            Not implemented yet.
+        Starts the OCR (optical character resolution) process on the uploaded documents.
+
+        Makes a background action for each page in the document.
         """
-        pass
+        if len(self.pages):
+            # We cannot OCR in place. It is too fragile.
+            for page in self.pages:
+                page.ocr_version = DAFile(page.attr_name("ocr_version"))
+                # psm=1 is the default which uses automatic text orientation and location detection.
+                # Appears to be the most accurate method.
+                page.ocr_status = page.ocr_version.make_ocr_pdf_in_background(
+                    page, psm=1
+                )
+
+    def ocr_ready(self) -> bool:
+        """
+        Returns:
+            True iff OCR process has finished on all pages. OCR is non-blocking, and assembly will work
+            even if OCR is not complete. Check this status if you want to wait to deliver a document until
+            OCR is complete.
+
+            Will return true (but log a warning) if OCR was never started on the documents.
+            That situation is likely a developer error, as you shouldn't wait for OCR if it never started
+        """
+        for page in self.pages:
+            if hasattr(page, "ocr_status") and not page.ocr_status.ready():
+                return False
+        return True
 
     def ocr_pages(self):
         """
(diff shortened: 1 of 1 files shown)


docassemble/AssemblyLine/al_document.py
```diff
@@ -1366,12 +1366,38 @@ def init(self, *pargs, **kwargs):
         if not hasattr(self, "starting_page"):
             self.start_page = 1
 
-    def add_numbers(self, starting_page=None) -> None:
```
nonprofittechy, line 1369: was `add_numbers` removed? Is there any evidence it is called somewhere?

docassemble/AssemblyLine/al_document.py
```diff
@@ -1393,25 +1419,38 @@ def ocr_pages(self):
...
+        *,
+        refresh: bool = False,
+        bates_prefix: str = "",
         pdfa: bool = False,
         add_page_numbers: bool = True,
         add_cover_page: bool = True,
         filename: str = None,
     ) -> DAFile:
-        if hasattr(self._cache, "_file"):
-            return self._cache._file
```
nonprofittechy, line 1403: This code was used to make sure the file was always fresh after an edit button is clicked. I can see that might not be compatible with OCR in the background though. Any thoughts on how to handle that need?

docassemble/AssemblyLine/al_document.py
```diff
@@ -1393,25 +1419,38 @@ def ocr_pages(self):
 
     def as_pdf(
         self,
-        prefix="",
```
nonprofittechy, line 1396: Agree that the new term is better, but I do worry that the old one might already be in use. Maybe add an alias, or do a quick GitHub search?

docassemble/AssemblyLine/al_document.py
```diff
@@ -1444,12 +1483,22 @@ def init(self, *pargs, **kwargs):
...
+            self.include_table_of_contents = True
+        if not hasattr(self, "include_exhibit_cover_pages"):
+            self.include_exhibit_cover_pages = True
+        if not hasattr(self, "bates_prefix"):
+            self.bates_prefix = ""
         self.object_type = ALExhibit
         self.complete_attribute = "complete"
 
     def as_pdf(
-        self, filename="file.pdf", pdfa: bool = False, add_cover_pages: bool = True
```
nonprofittechy, line 1452: Fairly sure the `add_cover_pages` parameter is in use.

docassemble/AssemblyLine/al_document.py
```diff
@@ -1489,38 +1537,31 @@ def _update_labels(self, auto_labeler: Callable = None) -> None:
 
     def ocr_ready(self) -> bool:
         """
-        Returns:
-            True iff OCR process has finished on all pages. OCR is non-blocking, and assembly will work
-            even if OCR is not complete. Check this status if you want to wait to deliver a document until
-            OCR is complete.
+        Combines the results of all the ALExhibit.ocr_ready calls.
+        See ALExhibi.ocr_ready for more details.
```
nonprofittechy, line 1541:
```suggestion
        Returns `True` iff all individual exhibit pages have been OCRed, or if the OCR process hasn't started.
```

docassemble/AssemblyLine/al_document.py
```diff
@@ -1595,6 +1645,12 @@ def has_overflow(self):
         """
         return False
 
+    def ocr_ready(self) -> bool:
+        """
+        See ALExhibitList.ocr_ready
```
nonprofittechy, line 1650:
```suggestion
        Returns `True` iff each individual exhibit has been OCRed, or if the OCR process was not started.
```

nonprofittechy (commented):
Overall looks great--just want to make sure you've checked for some of the parameters that have been renamed in at least our GitHub repos with a quick search.

docassemble/AssemblyLine/al_document.py
BryceStevenWilley: Github search only shows this file as having it.

As I mentioned above, IMO the only public API (at least the only API that we should actively use and expect people to use) is `ALExhibitDocument`, and maybe the fact that `ALExhibitList` is a `DAList`.

In general I was trying to reducing the duplicate of params. I moved params to objects-block / initialization only if I thought that they wouldn't change dynamically based on the interview.

docassemble/AssemblyLine/al_document.py
```diff
@@ -1489,38 +1537,31 @@ def _update_labels(self, auto_labeler: Callable = None) -> None:
...
 
-    def _update_page_numbers(self) -> None:
+    def _update_page_numbers(
+        self, starting_number: Optional[int] = None, toc_guess_pages: int = 1
+    ) -> None:
         """
         Update the `start_page` attribute of all exhibits so it reflects current position in the list + number of pages of each document.
         """
-        current_index = 1
+        toc_pages = toc_guess_pages if self.include_table_of_contents else 0
```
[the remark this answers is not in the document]
BryceStevenWilley: We already do this (and tested! It takes 28 attachments to overflow the ToC page, lol) [in the ALExhibitDocument](https://github.com/SuffolkLITLab/docassemble-AssemblyLine/pull/508/files#diff-db23667d014599821c0250a04d5ffbbcf327b89266e017c6f6304c30a19ff8efR1672), where we do have the table of contents document itself.  We can't do that here because we don't have the toc here.

nonprofittechy (approved):

PR #510, opened 2022-06-08
Distribute type information as a python package

BryceStevenWilley:
Following the instructions in https://mypy.readthedocs.io/en/latest/installed_packages.html#creating-pep-561-compatible-packages.

> packages that supply type information via type comments or annotations in the code should put a py.typed file in their package directory. 

The standard docassemble package data should take care of finding the file.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,1 +1,2 @@
 include README.md
+global-include *.typed


nonprofittechy (approved):
Nice catch! LGTM. Any issues caused by the zero byte file you can think of? It's ring a bell as potentially a problem but I could be getting mixed up with empty directories.

BryceStevenWilley:
I can't think of any issues with zero byte files. You're right, empty directories aren't added to git unless they have some files in them, but empty files are fine from what I know.

PR #512, opened 2022-06-09
Do not show `send_button_html()` if there are no enabled documents in a bundle

### LINKED ISSUE #503: send_button_html should return empty string if the bundle has no enabled documents

nonprofittechy:
Made a small refactor to logic in enabled_documents() and added an `is_enabled()` method that tracks the logic that was in `enabled_documents()`. This allows us to use a generator expression, so that we can return `True` as soon as we figure out at least one document is enabled.

Fix #503

### DIFF
--- a/docassemble/AssemblyLine/al_document.py
+++ b/docassemble/AssemblyLine/al_document.py
@@ -797,6 +797,25 @@ class ALDocument(DADict):
             input_width=input_width,
         )
 
+    def is_enabled(self, refresh=True):
+        """
+        A document can be considered "enabled" if:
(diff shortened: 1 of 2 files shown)


BryceStevenWilley (approved):
LGTM! Probably out of scope for this PR, but is there a reason we don't also want to change the behavior of `download_button_html` for no documents?

BryceStevenWilley:
> do you mean download_list_html? That method iterates directly over the list of documents, so it will already produce an empty string

It looks like it still generates an empty HTML table. I don't know if it's announced by screen readers or not (the screen reader I usually test with is acting up at the moment so I can't test it).

PR #518, opened 2022-06-10
Add `show_if` parameter to *_fields() methods

### LINKED ISSUE #273: Add a "show if" option to all _fields methods
It should be possible to add `show if` logic while using the `name_fields`, `address_fields`, etc. templatized field blocks.

nonprofittechy:
Fix #273

Also: [remove default mutable parameter](https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/) and do a little refactoring to reduce code repetition in those methods.

Test YAML:

```yaml
---
include:
  - assembly_line.yml
---
objects:
  - spouse: ALIndividual
---
sets:
  - users[0].name.first
  - spouse.name.first
question: |
  What is your name?
fields:
  - code: |
      users[0].name_fields()
  - Do you have a spouse?: has_spouse
    datatype: yesnoradio
  - code: |
      spouse.name_fields(show_if="has_spouse")
  - Do you have an address?: has_address
    datatype: yesnoradio
  - code: |
      users[0].address_fields(show_if="has_address")
  - Do you want to disclose your gender?: gender_disclosure
    choices:
      - I want to tell you my gender: say_more
      - I don't want to tell you my gender: dont_disclose
  - code: |
      users[0].gender_fields(show_if={"variable": "gender_disclosure", "is": "say_more"})
---
mandatory: True
question: |
  Your name is ${ users[0] }.
  
  % if has_spouse:
  Your spouse is ${ spouse }
  % endif
  ```

### DIFF
--- a/docassemble/AssemblyLine/al_general.py
+++ b/docassemble/AssemblyLine/al_general.py
@@ -583,7 +583,7 @@ class ALIndividual(Individual):
             show_if=show_if,
         )
 
-    def gender_fields(self, show_help=False, show_if: str = None):
+    def gender_fields(self, show_help=False, show_if: Union[str, Dict[str, str]] = None):
         """
         Return a standard gender input with "self described" option.
         """
--- a/docassemble/AssemblyLine/al_general.py
+++ b/docassemble/AssemblyLine/al_general.py
@@ -583,7 +583,9 @@ class ALIndividual(Individual):
             show_if=show_if,
         )
 
-    def gender_fields(self, show_help=False, show_if: Union[str, Dict[str, str]] = None):
+    def gender_fields(
+        self, show_help=False, show_if: Union[str, Dict[str, str]] = None
+    ):
         """
         Return a standard gender input with "self described" option.
         """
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: docassemble/AssemblyLine/al_general.py [hunk 1]
```diff
@@ -558,9 +580,10 @@ def address_fields(
             country_code=country_code,
             default_state=default_state,
             show_country=show_country,
+            show_if=show_if,
         )
 
-    def gender_fields(self, show_help=False):
+    def gender_fields(self, show_help=False, show_if: str = None):
```
BryceStevenWilley, line 586:
```suggestion
    def gender_fields(self, show_help=False, show_if: Union[str, Dict[str, str]] = None):
```

BryceStevenWilley (approved):
Overall LGTM. The test interview is kinda confusing, since with more than just a single name / address input, it's hard to tell what exactly is being filled in, but there are workarounds (abet a little cumbersome) to making it more clear, like manually having a showif'd `note`, or changing the `name_labels` for a person you know will appear on that screen, and I can't think of any simple way to improve it through this interface.

PR #531, opened 2022-06-29
ALDocumentBundles need `is_enabled` too

BryceStevenWilley:
Fixes an error caused by an error when trying to use nested ALDocumentBundles.


nonprofittechy (approved):
Good catch

PR #542, opened 2022-07-18
First check if the bundle's enabled attribute is on

### LINKED ISSUE #538: Bundles not respecting enabled attribute
**Describe the bug**
Bug introduced in https://github.com/SuffolkLITLab/docassemble-AssemblyLine/pull/531, document bundles don't look at their own `enabled` attribute anymore.

**To Reproduce**
1. Make a document bundle with one document:
```yaml
objects: 
  - my_bundle: ALDocumentBundle.using(title="hi", elements=[my_doc], ...)
---
objects:
  - my_doc: ALDocument.using(title="document", ...)
```
2. Disable the bundle, but enable the doc.
```yaml
code: |
  my_doc.enabled = True
  my_bundle.enabled = False
---
```
3. My bundle will still be enabled / show up if you do something like `my_bundle.download_html_list`, or put it in another bundle and do `that_bundle.download_html_list`. 

**Expected behavior**
The bundle should only be enabled if the bundle itself is enabled and it has enabled documents itself.

 - AssemblyLine Version: 2.13.0

BryceStevenWilley:
Fixes #538

An example playground to play around with, simply upload a `my_pdf.pdf` file to that project:

```yaml
---
include:
---
objects:
  - attachment: ALStaticDocument.using(title="Attachment", filename="my_pdf.pdf", has_addendum=False)
  - instructions: ALStaticDocument.using(title="Instructions", filename="my_pdf.pdf", has_addendum=False)
  - additional: ALStaticDocument.using(title="Additional", filename="my_pdf.pdf", has_addendum=False)
---
code: |
  attachment.enabled = True
  instructions.enabled = True
  additional.enabled = True
---
objects:
  - al_user_bundle: ALDocumentBundle.using(elements=[instructions, bundle], filename="appearance.pdf", title="Download all forms")
---
objects:
  - bundle: ALDocumentBundle.using(elements=[attachment, additional], filename="appearance.pdf", title="Bundle", enabled=False)
---
mandatory: True
question: Test first screen
subquestion: otherwise DA might 504 during development
continue button field: fake_field
---
mandatory: True
question: Hi
subquestion: |
  ${ al_user_bundle.download_list_html() }
```

### DIFF
--- a/docassemble/AssemblyLine/al_document.py
+++ b/docassemble/AssemblyLine/al_document.py
@@ -1444,19 +1444,25 @@ class ALDocumentBundle(DAList):
                 attachments=self.as_pdf_list(key=key),
                 **kwargs,
             )
+    
+    def _is_self_enabled(self, refresh=True) -> bool:
+        """The same as ALDocument.is_enabled"""
+        if hasattr(self, "always_enabled") and self.always_enabled:
+            return True
+        if hasattr(self.cache, "enabled"):
+            return self.cache.enabled
+        if refresh:
+            self.cache.enabled = self.enabled
+            if hasattr(self, "enabled"):
+                del self.enabled
+            return self.cache.enabled
+        else:
+            return self.enabled
+      
 
     def is_enabled(self, refresh=True) -> bool:
         """Returns true if the bundle itself is enabled, and it has at least one enabled child document"""
-        self_enabled = (hasattr(self, "always_enabled") and self.always_enabled) or \
-            (hasattr(self.cache, "enabled") and self.cache.enabled)
-        if not self_enabled:
-            if refresh:
-                self.cache.enabled = self.enabled
-                self_enabled = self_enabled or self.cache.enabled
-                if hasattr(self, "enabled"):
-                    del self.enabled
-            else:
-                self_enabled = self_enabled or self.enabled
+        self_enabled = self._is_self_enabled(refresh=refresh)
         return self_enabled and self.has_enabled_documents(refresh=refresh)
 
 
--- a/docassemble/AssemblyLine/al_document.py
+++ b/docassemble/AssemblyLine/al_document.py
@@ -1445,8 +1445,19 @@ class ALDocumentBundle(DAList):
                 **kwargs,
             )
 
-    def is_enabled(self, refresh=True):
-        return self.has_enabled_documents(refresh=refresh)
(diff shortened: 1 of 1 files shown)


docassemble/AssemblyLine/al_document.py
```diff
@@ -1445,8 +1445,19 @@ def send_email(
...
 
-    def is_enabled(self, refresh=True):
-        return self.has_enabled_documents(refresh=refresh)
+    def is_enabled(self, refresh=True) -> bool:
+        """Returns true if the bundle itself is enabled, and it has at least one enabled child document"""
+        self_enabled = (hasattr(self, "always_enabled") and self.always_enabled) or \
+            (hasattr(self.cache, "enabled") and self.cache.enabled)
+        if not self_enabled:
+            if refresh:
+                self.cache.enabled = self.enabled
```
nonprofittechy, line 1454: How does this work if:
- always_enabled is False
- the user changes an option that makes the bundle NOT enabled?

BryceStevenWilley: I should have moved the second half of the `or` statement inside `if not self_enabled`, but either way, if `always_enabled is False` nothing will turn it on.

And yeah, on refresh it should reconsidered everything but `always_enabled` (if it's not in the cache).

The logic was just taken from `ALDocument.is_enabled`. It's not in the same object hierarchy, so I can't just call it.

nonprofittechy: I think `always_enabled` is only supposed to be True if `enabled` is set to true inside the `init` method, i.e., in the `.using()` method in the `objects` block. So in most of the interesting situations it will start out `False`.

I just am making sure this logic works if `always_enabled` starts out `False`, someone enables it (so the .cache.enabled property is at least temporarily true) and then they change their mind.

The answer might be that the `cache` is the only place we're checking and that's erased on screen refresh. If so that's good--and I think it's working correctly in `ALDocument.is_enabled()`

BryceStevenWilley: Misunderstood how always_enabled was supposed to work. The above should just be the same ALDocument, I'm not trying to make it any different, but refactored it a bit to make that clearer.

nonprofittechy (approved):
LGTM, the new expanded logic is easier to follow. Thanks!

PR #549, opened 2022-08-05
DA 1.4.0 compatibility

BryceStevenWilley:
Now we're on python 3.10, there's a `zoneinfo` directly in python, and the `backports.zoneinfo` module doesn't exist.

Adds a try-except for the import, and additionally install the `backports.zoneinfo` package when AL is installed (without which, it will error on a fresh DA instance).

### DIFF
--- a/docassemble/AssemblyLine/sessions.py
+++ b/docassemble/AssemblyLine/sessions.py
@@ -42,9 +42,9 @@ from .al_document import (
 import json
 import os
 try:
-  import zoneinfo
+  import zoneinfo # type: ignore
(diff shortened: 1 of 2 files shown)


BryceStevenWilley:
ALKiln test failing because it's taking an incredibly long time for anything to happen on dev/test servers. For some reason the interviews spin for 15-20 seconds before actually starting to process the next page (when looking at the debug timings). Not sure why, as it's not happening on a local DA server. Can't tell where to debug without more info, but either way, things do actually pass.

PR #555, opened 2022-08-11
Fix no address

BryceStevenWilley:
`hide if` will hide the field if the variable isn't present.

Also adds a simple ALKiln test to look for things like this again, just in the question library versions of address, no address, and contact info. ALKiln test passes locally.


nonprofittechy:
Thanks, I wonder if that was a changed default!

PR #556, opened 2022-08-16
Don't add left-padding to the first footer child

BryceStevenWilley:
Prevents the last link name ("feedback" by default) from being cut off by not
adding more padding to the left of the first child.

Notably, let's us show all 3 of the footer links in full on smaller mobile
devices, specifically iPhone 5/SE (the smallest I can simulate in Firefox dev
tools), at 320 px wide. Will show all three up above 308px.

Doesn't change anything on larger screens, since `text-align: right` is set in the footer.

### Before patch, iPhone 5 (simulated)

Screenshot from 2022-08-16 14-34-28

### After patch, iPhone 5 (simulated)

Screenshot from 2022-08-16 14-34-15


nonprofittechy (approved):
Looks good! Just to check to see if should have a media query--it looks good on big screens still, too. Thanks!

image

Issue #109, opened 2021-04-28
Anywhere that we say "you" we should consider how to account for forms that are always used by an attorney


nonprofittechy:
This should be easily solvable by using `set_info()` https://docassemble.org/docs/functions.html#set_info

nonprofittechy:
The easy part is rewriting the text. The hard part is figuring out the standard.

nonprofittechy:
This is a live issue with the https://github.com/SuffolkLITLab/docassemble-IlFeeWaiver form where you can file on behalf of someone else.

I think there are two separate issues:

1. Make all of the questions that say "you" responsive to the value of `get_info('user')`. Set a default for that.
2. Come up with a new standard variable name to represent the person filling in the form. In ILFeeWaiver, we are using the variable "signer" which I think works pretty well but I can imagine a lot of forms where the variable name will get re-used for a different meaning. Maybe something long like "question_respondent", "person_answering_interview", "form_filler", or just "interview_user" or perhaps prefixed "al_interview_user" to indicate it's about the interview, not the actual document is best.

I think we can safely start refactoring the ql_baseline questions but we should also make sure to update all of the translations that we now have into various languages so those aren't instantly broken.

BryceStevenWilley:
> Make all of the questions that say "you" responsive to the value of get_info('user'). Set a default for that.

So I understand how we would be using `set_info`, to solve 1, we would have to:
* remove the questions that specifically use `users[0]` as the target object
* use the object methods for other ALIndividual questions, i.e. instead of ["What is ${ users[i] }'s address"](https://github.com/SuffolkLITLab/docassemble-AssemblyLine/blob/f4c74de8dcc945e323ce1eb8089fd907269fe432/docassemble/AssemblyLine/data/questions/ql_baseline.yml#L479), we'd do:
  ```mako
  What is ${ users[i].possesive('address') }?
  ```
* call `set_info` early in the interview if and only if the user says they're filling it out for themselves (this is also 3, right? just the variable to store that question in?)

>  Set a default for that.

I'm not sure what you mean there. Is it the default for if the user is the person answering questions (i.e. true or false)? That should be what we've been assuming in the current interviews, that `users[0]` is the person answering questions, right?

nonprofittechy:
I don't think we'd need to re-target any existing questions, just change the dynamic language, but it's worth figuring out the clearest/least disruptive change.

[...]

PR #557, opened 2022-08-23
Add question for your birthdate

BryceStevenWilley:
Prevents the too general "What is Bob Miller's Birthdate" from being asked by default.

### DIFF
--- a/docassemble/AssemblyLine/data/questions/ql_baseline.yml
+++ b/docassemble/AssemblyLine/data/questions/ql_baseline.yml
@@ -424,7 +424,7 @@ fields:
 ---
 id: you birthdate question
 question: |
-  When were your born?
+  When were you born?
 fields:
   - Birthdate: users[0].birthdate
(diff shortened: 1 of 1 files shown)


docassemble/AssemblyLine/data/questions/ql_baseline.yml
```diff
@@ -422,6 +422,13 @@ fields:
   - Birthdate: x.birthdate
     datatype: BirthDate
 ---
+id: you birthdate question
+question: |
+  When were your born?
```
nonprofittechy, line 427:
```suggestion
  When were you born?
```

nonprofittechy: As we add more questions with direct address, we should really work on #109. But I do think this is a safe change in keeping with our other direct address questions

nonprofittechy (approved):

PR #559, opened 2022-08-25
Save progress and form title at regular intervals, small interview list cleanups

### LINKED ISSUE #522: Save interview title and progress to SQL metadata at regular intervals?
While working on #237, it would be nice to:

1. View a descriptive title for existing sessions and
2. Be able to view the progress of existing sessions

Parsing this from the interview is too slow. Ideally, we would save this information as we go to an unencrypted format, such as the jsonb "metadata" table that we already use to add metadata to answer sets.

But: when would we save it? On every screen load, in an `initial` block? At a specific save points in the interview? (that would require editing existing forms).

This is mostly a performance question. Using an `initial` block would work great and be backwards compatible, but it might be too slow. While we could save the name at a specified save point, that wouldn't really work for setting progress.

nonprofittechy:
Fix #522
Also, some small cleanups to interview list page--delete confirmation, only show next/delete when relevant, make the clickable link be the name the user chose rather than the generic form title.

Progress is taken from set_progress(). Title comes from a template or string that can be overridden:

`al_sessions_interview_title`. The default contents of this template are the user v. other_parties, falling back to just the user, falling back to just the form's title.

I've taken a cautious approach to this feature. It's behind a configuration flag:

```yaml
assembly line:
  update answer set metadata: true  
```

To reduce database load (maybe unnecessary) it saves the progress and author title to the database every 3 steps or every step when the interview is > 99 progress.

So far it seems to work pretty well. We might want to tweak the "progress" bubbles a little more.

image

So far I did not notice any slowdown at all, which makes sense since it's just a couple of extra local DB queries. But I'm not totally sure how to measure the performance impact.


docassemble/AssemblyLine/data/questions/al_saved_sessions.yml
```diff
@@ -4,6 +4,34 @@ code: |
   # Create a placeholder value to avoid playground errors
   al_sessions_snapshot_results = DAEmpty() 
 ---
+initial: True
+code: |
+  ########### Save title and progress to session list ##################
+  if get_config("assembly line",{}).get("update answer set metadata"):
+    # To limit impact on DB, save progress every 3 steps or when progress is 100 +- 1
+    if get_progress() > 99 or int(all_variables(include_internal=True).get('_internal',{}).get('steps',1)) % 3 == 1:
```
BryceStevenWilley, line 12:
```suggestion
    if get_progress() > 99 or int(_internal.get('steps',1)) % 3 == 1:
```
Should be the same thing, it returns the same info in some demo interviews.

BryceStevenWilley (approved):
LGTM, I don't think you need `all_variables` to get the steps, but everything else looks solid

PR #566, opened 2022-09-13
Rename "docket number" to case number

### LINKED ISSUE #564: Rename "Docket number" (the question text) to "Case number"
Docket number is a Massachusetts-specific term. We should move that into the ALMassachusetts package.

nonprofittechy:
Fix #564. Docket number, as far as we can tell with the dozen or so states we've interacted with, is a pretty Massachusetts-specific term. The more general term is "case number". This changes the phrasing of "Docket number" to be "case number" so that people in other states can get better interviews out of the box.

See https://github.com/SuffolkLITLab/docassemble-ALMassachusetts/issues/7 which will move the docket number phrasing to ALMassachusetts, where it is the less confusing term

### DIFF
--- a/docassemble/AssemblyLine/data/questions/ql_baseline.yml
+++ b/docassemble/AssemblyLine/data/questions/ql_baseline.yml
@@ -10,6 +10,7 @@ objects:
   - other_parties: ALPeopleList
   - witnesses: ALPeopleList.using(ask_number=True)
   - docket_numbers: DAList
+  - case_numbers: DAList
   - caregivers: ALPeopleList
   - guardians_ad_litem: ALPeopleList
   - attorneys: ALPeopleList
@@ -27,8 +28,6 @@ objects:
   - courts: DAList.using(auto_gather=False, target_number=1, gathered=True)
   - trial_court_address: Address
   - appeals_court_address: Address
-  
-  
 ---
 id: basic questions intro screen
 decoration: form-lineal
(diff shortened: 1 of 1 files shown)


BryceStevenWilley (commented):
+1 to changing the wording of the question, not as sure about adding an additional `case_numbers` variable though. Some questions that would ease my mind there:

* are there any cases where we have forms interacting across jurisdictions, where one set of forms would use docket number and the other would use case number? Splitting the same information into two separate variables would make things difficult there
* the EFileProxy will need to change to using `case_number` in addition to docket number. It seems rare enough that both would be set at the same time, so I'm inclined to just return an error on interviews where that's the case. Does that sound like the right solution?

nonprofittechy:
> +1 to changing the wording of the question, not as sure about adding an additional `case_numbers` variable though. Some questions that would ease my mind there:
> 
>     * are there any cases where we have forms interacting across jurisdictions, where one set of forms would use docket number and the other would use case number? Splitting the same information into two separate variables would make things difficult there
> 
>     * the EFileProxy will need to change to using `case_number` in addition to docket number. It seems rare enough that both would be set at the same time, so I'm inclined to just return an error on interviews where that's the case. Does that sound like the right solution?

Didn't realize the e-filing stuff depended on `docket_number`, so I don't have a problem removing the `case_number` variable and just making it a AssemblyLine ism.

PR #569, opened 2022-09-15
Fix session list on small screens

### LINKED ISSUE #568: Set .text-break on columns in sessions.py
See https://getbootstrap.com/docs/5.2/utilities/text/#word-break

Currently, on mobile, even relatively short interview names will overflow messily into the next column since we use a 4-column layout.
Maybe we should also consider stacking the actions vertically on small viewports?

nonprofittechy:
Currently on small screens, if the title of the form uses PascalCase, it will bleed into the second column.
This change:
- Inserts an invisible space (zero width space) into PascalCase words so there is a natural break point
- Turns on the bootstrap text-break utility so that further breaks can happen even in the middle of a word instead of breaking the display

Fix #568

### DIFF
--- a/docassemble/AssemblyLine/sessions.py
+++ b/docassemble/AssemblyLine/sessions.py
@@ -41,6 +41,7 @@ from .al_document import (
 )
 import json
 import os
+import re
 
 try:
     import zoneinfo  # type: ignore
@@ -447,7 +448,7 @@ def interview_list_html(
     if not answers:
         return ""
 
-    table = '<div class="table-responsive"><table class="table table-striped al-saved-answer-table">'
+    table = '<div class="table-responsive"><table class="table table-striped al-saved-answer-table text-break">'
     table += f"""
     <thead>
       <th scope="col">
@@ -473,7 +474,7 @@ def interview_list_html(
             """
         else:
             table += f"""
-            <td><a href="{ url_action(load_action, i=answer.get("filename"), session=answer.get("key")) }">{ nice_interview_subtitle(answer) or nice_interview_title(answer) }</a></td>
+            <td class="text-break"><a href="{ url_action(load_action, i=answer.get("filename"), session=answer.get("key")) }">{ nice_interview_subtitle(answer) or nice_interview_title(answer) }</a></td>
             """
         table += f"""
         <td>{ as_datetime(answer.get("modtime")) }</td>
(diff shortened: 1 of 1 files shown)


docassemble/AssemblyLine/sessions.py
```diff
@@ -518,21 +519,31 @@ def nice_interview_title(
...
 
 
+def pascal_to_zwspace(text: str) -> str:
+    """
+    Insert a zero-width space into words that are PascalCased to help
+    with word breaks on small viewports.
+    """
+    re_outer = re.compile(r"([^A-Z ])([A-Z])")
+    re_inner = re.compile(r"(?<!^)([A-Z])([^A-Z])")
+    return re_outer.sub(r"\1​\2", re_inner.sub(r"​\1\2", text))
```
plocket, line 529: I'm not sure what I'm doing wrong, but I can't get this to work: https://replit.com/@plocket/scratchpad#main.py. Also, I don't see spaces in here, but strange red dots appeared in the replit editor. When I add actual spaces, it seems to work fine.

```suggestion
    re_outer = re.compile(r"([^A-Z ])([A-Z])")
    re_inner = re.compile(r"(?<!^)([A-Z])([^A-Z])")
    return re_outer.sub(r"\1 \2", re_inner.sub(r" \1\2", text))
```

plocket (commented):
It may need some changes, but you may know something I don't know, so I don't want to block this.

docassemble/AssemblyLine/sessions.py
```diff
@@ -518,21 +519,31 @@ def nice_interview_title(
...
+
 def nice_interview_subtitle(answer: Dict[str, str], exclude_identical=True):
     """
     Return first defined of the "title" metadata, the "auto_title" metadata, or empty string.
 
     If exclude_identical, return empty string when title is the same as the subtitle.
     """
     if answer.get("title"):
-        return answer.get("title")
+        return pascal_to_zwspace(answer["title"])
```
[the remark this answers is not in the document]
nonprofittechy: This only affects where line breaks appear if they are on a small screen. That's the purpose of a zero width space.

docassemble/AssemblyLine/sessions.py
nonprofittechy: We don't want a real space. The strange red dot is a zero width (invisible) space that is only used to control where line breaks are inserted.

PR #576, opened 2022-09-20
Don't turn on address autocomplete if not enabled

BryceStevenWilley:
Checks if the google maps API is present on the server, and if it isn't, does not use address autocomplete in `ALAddress.address_fields()`.

Not necessary for our servers, but helpful when running on local servers, and for users who might not wath Google autocomplete enabled.


nonprofittechy (approved):
Looks fine. This won't catch problems when API keyis present but invalid or when billing isn't enabled, but if this improves local install experience that works for me.

PR #579, opened 2022-09-22
Format user-entered phone numbers in `phone_numbers()` method

nonprofittechy:
Sample interview:

```
---
include:
  - assembly_line.yml
---
mandatory: True
code: |
  users[0].phone_number
---
mandatory: True
question: |
subquestion: |
  ${ users[0].phone_numbers() }
  
  ${ phone_number_formatted(users[0].mobile_number, country=None) }
```

### DIFF
--- a/docassemble/AssemblyLine/al_general.py
+++ b/docassemble/AssemblyLine/al_general.py
@@ -11,6 +11,7 @@ from docassemble.base.util import (
     url_action,
     get_config,
     phone_number_is_valid,
+    phone_number_formatted,
     validation_error,
     DAWeb,
     get_config,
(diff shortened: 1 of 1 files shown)


docassemble/AssemblyLine/al_general.py
```diff
@@ -479,12 +480,30 @@ def signature_if_final(self, i: str) -> Union[DAFile, str]:
...
-            nums.append({self.mobile_number: "cell"})
+            try:
+                nums.append(
+                    {
+                        phone_number_formatted(
+                            self.mobile_number, country=country
+                        ): "cell"
+                    }
+                )
+            except:
```
BryceStevenWilley, line 494: Did you run into an exceptions when doing this, or are you just being extra cautious? Looking through `phone_number_formatted`, it doesn't look like there's anything that would throw, I'm okay with either, just curious.

nonprofittechy: I think the function used to return an exception if the phone number didn't match the country code but it looks like it doesn't anymore. So might be redundant now.

PR #586, opened 2022-09-29
Close #563 Add send email table row and change styling to align buttons

### LINKED ISSUE #563: Improve visual clarity and consistency of the `send_button_html` and `download_list_html()` elements
The current buttons are a little messy, especially if you have two separate bundles:

image

The `download_list_html()` buttons and the `send_button_html()` buttons:

1. Are not on the same grid
2. Are not clearly separated or grouped by a proper amount of white space
3. Have a confusing visual hierarchy--the `h5` that labels the `send_button_html()` output is larger than the name of the document, and the document title is bold while the "send as email" heading is fairly light weight
4. Are not clearly grouped together as one logical unit (unifying the send button with the list of documents)

See also #562

plocket:
Closes #563, alignment of document download table buttons vs. document email section parts, plus other alignment and styling issues. [I tried to limit the changes to the email section as people are using that styled as-is right now.] Here are some notes from our deep dive doc as well:

* We’re going to add a new parameter to download_list_html() to include the send email button. But it will be a copy of the feature - we won’t delete the existing send_button_html() method
* We need to fix the send_button_html() method to remove the H4 and replace with H2 class="h4"
* Maybe we want to right align the send button of the separate email section, but we need to check with Matt to see if it will mess up his styles
* We want to remove the big gutter between the buttons and the text

ACCESSIBILITY QUESTION: Giving the separate email section an `h2` tag assumes that it's always the second level header after the title at the top of the page. I can easily see it being under a different `h2` - the name of document they're sending. Basically `h2`, or any level of header, seems like an assumption I'm not sure we can make if we want to align with accessibility. Thoughts?

Screenshots of the tables and send email section with different widths:

### Narrower

narrower_screen

---

### Wider

wider_screen

### DIFF
--- a/docassemble/AssemblyLine/al_document.py
+++ b/docassemble/AssemblyLine/al_document.py
@@ -95,23 +95,22 @@ def html_safe_str(the_string: str) -> str:
     return re.sub(r"[^A-Za-z0-9]+", "_", the_string)
 
 
-# def table_row( title, view_file:DAFile, download_file:DAFile=None, view_icon:str="eye", download_icon:str="download") -> str:
 def table_row(title: str, button_htmls: List[str] = []) -> str:
     """
     Uses the provided title and list of button html strings to
     return the row of an AL document-styled table in HTML format.
     """
     html = (
-        f"\n\t<tr>"
-        # '\n\t\t<td><i class="fas fa-file"></i>&nbsp;&nbsp;</td>'
-        # TODO: Need to replace with proper CSS
-        f'\n\t\t<td class="al_doc_title"><strong>{title}</strong></td>'
-        f'\n\t\t<td class="al_buttons">'
+        f'\n\t<div class="row al_doc_table_row">''
+        f'\n\t\t<div class="col col-6 al_doc_title">{title}</div>'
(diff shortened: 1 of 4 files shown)


BryceStevenWilley:
> ACCESSIBILITY QUESTION: Giving the separate email section an h2 tag assumes that it's always the second level header after the title at the top of the page. I can easily see it being under a different h2 - the name of document they're sending. Basically h2, or any level of header, seems like an assumption I'm not sure we can make if we want to align with accessibility. Thoughts?

There shouldn't be any accessibility issues with `<h2>`. The accessibility rules with headers are mostly:
* there should only be a single h1
* you can't jump down to a lower header level. i.e. h1 -> h3 can confuse some users. However, you can jump back up: h4 followed by h2 is fine because you're just starting another h2: you don't need to add an h3 between the h4 and the h2.

So I think an h2 should always be fine: we will never be jumping down, as there's always at least one h1 on the page, and it's fine for an h2 to be followed by another h2.

nonprofittechy:
Is the big gutter on wide screens unavoidable?

PR #591, opened 2022-10-03
Improve usability of signature question

### LINKED ISSUE #585: Usability issue providing a signature
```
id: signature choice
decoration: file-signature
question: |
  Sign your forms
subquestion: |
  Your forms are almost ready. You need to sign them first.
  
  Use the mouse or touchpad on your computer or
  sign with your finger on a smartphone.
```

Observed a user read the instructions and then try to sign on their screen next to the words "SIgn your forms" which make a horizontal line.

nonprofittechy:
Fix #585

This just makes the changes we discussed on our call today. It's a pretty simple fix but touches most of our interviews so I welcome additional feedback I missed on the call.

image

I added a new flag variable, `al_form_requires_digital_signature` which can be used to add a third option that's currently missing: "sign after I print". When this is chosen, the `x.signature` attribute will be defined as `DAEmpty()`

### DIFF
--- a/docassemble/AssemblyLine/data/questions/al_code.yml
+++ b/docassemble/AssemblyLine/data/questions/al_code.yml
@@ -126,7 +126,10 @@ code: |
     # 3. Used the QR code
     saw_signature_choice
     # Branch 1: PC
-    if signature_choice =='this device':
+    if signature_choice == "sign_after_printing":
+      for signature in signature_fields:
+        define(signature, DAEmpty())
+    elif signature_choice in ['this device', 'this_device']:
       for signature in signature_fields:
         value(signature)    
     # Check for Branch 2 or 3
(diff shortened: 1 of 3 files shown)


nonprofittechy:
I'm ambivalent about the extra value of the images in the radio buttons--makes it a little busy to my eye. but open to feedback

BryceStevenWilley:
Putting the words on the choices themselves made me realize that we should change `started_on_phone` to `device().is_touch_capable`, which determines if it has a touch capable, as opposed to if it's mobile (which specifically excludes tablets).

I'd also suggest changing "Give me a link so I can sign on my phone with my finger" to just "Sign on my phone with my finger", or "Give me a link so I can sign on my phone" so it fits on one line. I don't think whole sentence is needed, as we give a lot of information on the next screen.

nonprofittechy:
Good suggestions, will push them up in a second. Here's the revised look

image

image

BryceStevenWilley (approved):
LGTM! I did kinda like the icons, but not enough to convince you to put them back 😄

PR #594, opened 2022-10-05
Close #563 Add send email table row and change styling to align buttons

### LINKED ISSUE #177: Make the ALDocument send_button_html CSS use classes and not IDs
Current code will not work for more than one bundle, or a bundle that doesn't match our al_court and al_user bundle names

### LINKED ISSUE #562: `send_button_html` should not use H5
Currently, `send_button_html()` uses an H5 element. This is unlikely to be the right H level.

Ideas:
1. Use a `span` instead?
2. Use `h2` (which is most likely to be correct for screen readers) with a style that makes it look like the current H5...although current H5 is a little messy anyway.
3. Reconsider the layout entirely, perhaps with a Bootstrap card: https://getbootstrap.com/docs/5.2/components/card/

Probably the right "quick" fix is to remove H5 and just use `<h2 class="h5">` as we've been doing in other places to workaround Docassemble's broken heading hierarchy

### LINKED ISSUE #563: Improve visual clarity and consistency of the `send_button_html` and `download_list_html()` elements

plocket:
Closes #563, Closes #562, alignment of document download table buttons vs. document email section parts, plus other alignment and styling issues. I tried to limit the changes to the email section as people are using that styled as-is right now. Here are some notes from our deep dive doc as well:

* Add a new parameter to download_list_html() to include a send email button row. But it will be a copy of the feature - we won’t delete the existing send_button_html() method.
* Fix the send_button_html() method html strings to remove the H5, though keep the styling
* Right align the send button of the separate email section,
* Remove the big gutter between the buttons and the text

Remade this PR because the pfs/black changes on the old branch caused a conflict with my updates.

Also closes #177 I believe.

### DIFF
--- a/docassemble/AssemblyLine/al_document.py
+++ b/docassemble/AssemblyLine/al_document.py
@@ -1432,7 +1432,7 @@ class ALDocumentBundle(DAList):
         the bundle to the specified email address.
 
         Optionally, display a checkbox that allows someone to decide whether or not to
-        include an editable (Word) copy of the file, if it is available.
+        include an editable (Word) copy of the file, if and only if it is available.
         """
         if not self.has_enabled_documents():
             return ""  # Don't let people email an empty set of documents
--- a/docassemble/AssemblyLine/al_document.py
+++ b/docassemble/AssemblyLine/al_document.py
@@ -1280,7 +1280,7 @@ class ALDocumentBundle(DAList):
                 icon=download_icon,
                 color="primary",
                 size="md",
-                classname="al_download",
+                classname="al_download al_button",
             )
             if view and doc.as_pdf().url_for().endswith(".pdf"):
                 doc_view_button = action_button_html(
@@ -1293,7 +1293,7 @@ class ALDocumentBundle(DAList):
                     icon=view_icon,
                     color="secondary",
                     size="md",
-                    classname="al_view",
+                    classname="al_view al_button",
                 )
                 buttons = [doc_view_button, doc_download_button]
             else:
@@ -1313,7 +1313,7 @@ class ALDocumentBundle(DAList):
                 icon=zip_icon,
                 color="primary",
                 size="md",
-                classname="al_zip",
+                classname="al_zip al_button",
             )
             html += table_row(zip.title, zip_button)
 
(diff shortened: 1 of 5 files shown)


docassemble/AssemblyLine/al_document.py
```diff
@@ -1360,7 +1408,7 @@ def send_button_html(
         the bundle to the specified email address.
 
         Optionally, display a checkbox that allows someone to decide whether or not to
-        include an editable (Word) copy of the file, iff it is available.
+        include an editable (Word) copy of the file, if it is available.
```
nonprofittechy, line 1411:
```suggestion
        include an editable (Word) copy of the file, if and only if it is available.
```
FYI, maybe I will avoid this shorthand but iff = if and only if--not the same as "if"

nonprofittechy (changes requested):
This is a tricky task--and thanks for all of your work on it so far!

I like the look at various sizes, but on a standard mobile device, the text in the button is now wrapping, which is not very visually attractive:

image

Maybe we can adjust the gutters to fix that? We should get it right at least on an iPhone 11 or newer.

On an iPhone 5, we have horizontal scrolling, which is also not ideal.

nonprofittechy:
One idea -- if it's just too hard to fit on mobile, maybe stack the form title and the view/download buttons vertically on mobile devices? We do need to really get that visual look right because it's a large proportion of our uses - 55%

nonprofittechy:
FYI, to avoid merge conflicts--pull this branch into your playground before making more changes.

plocket:
Not sure what we want to do with the pre-existing email elements because we said didn't want to change those too much, but this is what stuff looks like now at less than 576px (shown on Galaxy Fold):

Screen Shot 2023-01-17 at 6 02 55 PM

I put validation errors below the email input field instead of next to it:

Screen Shot 2023-01-17 at 6 20 54 PM

That leaves us with this problem on bigger screens that I'm still trying to figure out:

Screen Shot 2023-01-18 at 12 41 16 PM

Just fyi, I keep having problems with the browser inspector trying out devices - under some circumstances I get weird behavior, like needing to scroll when in phone mode or the elements becoming narrower than the screen. You might be running into this too.

Some width stuff has to be somewhat home-brewed (from what I can tell so far) because we have to handle more than just column widths. For example, aligning doc titles left instead of right. If someone knows how to do this through bootstrap, give a shout.

BryceStevenWilley:
Try merging with main, I fixed that error earlier last month, Github made a change in the underlying container I think

plocket:
Since this issue is about the download table, it may be worth leaving the isolated email component out of this PR and dealing with it in another issue.

BryceStevenWilley:
I'm not sure how you merged, but it unfortunately has muddled the diff; I'm not sure why all of the other changes are showing up in it now. You could revert the changes, and maybe try rebasing on main instead? You can make a copy of the branch (`git checkout -b doc_table_alignment_dupe`) before trying, just so you don't lose info

plocket:
Tried to rebase in GitHub, but it said there were conflicts. I'll try anyway.

plocket:
Seemed to work when rebasing locally. If the tests all pass, is it ok with folks if I force-push to this branch and this PR so that we can keep the PR comment history?

Issue #181, opened 2021-06-07
Cleanup use of the `user_role`, `user_ask_role`, `user_started_case`, `typical role`, `user_detailed_role` and possibly other signals


nonprofittechy:
# Problem statement
figure out how to represent information about both whether this is a pleading that starts a case or is continuing a case; whether the party is the person who initiated the case; and the specific name of their role in the case. (Role meaning which side of the "v" in the case caption they are on and the name of that side of the "v" in the particular case).

Most forms ONLY care if the user started the case or is responding to a case someone else started. In some forms, the "starter" is called Plaintiff, and others it is called "Petitioner" and the person responding is called "Defendant" or "Respondent." But the interview does not need to distinguish. It is enough to set the value of both "plaintiffs" and "petitioners" and let the author use the one that is appropriate for the case. That is our current practice.

A small subset of forms are general purpose and are filed in existing cases that can be of the "petitioner" or "plaintiff" variety and the interview author does not know in advance. An even smaller subset require the end-user of the form to identify whether they are one of "petitioner", "plaintiff", "respondent", or "defendant". We want to be able to ask a question that gets to that very specific role for the small number of forms that need it, but we do not want to force end users to get so specific if it's not necessary for the form. They are likely to lack the context to be able to answer the specific choice among those 4 possibilities. We should only force the choice on them when there is no better option.

# Goal
Use a minimum number of clear variable names.

# What is going on now

We have several variables that appear as though they might overlap. We want to enable interview authors to make forms of the appropriate level of detail for their needs but also that require the fewest clicks/customization to do so that is feasible.

# Possible scope

1. Identify every place each of these variables are used.
2. Identify the purpose behind each use.
3. Figure out if any of the uses could be functionally combined.
4. Identify needs that are not met by the current usage.
5. Identify the risk and benefits of changing existing code and figure out the most appropriate solution.

# Existing variables (more info to be gathered, w/ line numbers)

## `al_form_type`

nonprofittechy:
See also 

* Related Weaver "meta" issue: https://github.com/SuffolkLITLab/docassemble-ALWeaver/issues/408
* Add a flag for `user_detailed_role` https://github.com/SuffolkLITLab/docassemble-ALWeaver/issues/415
* How to document the meaning of the various variables: https://github.com/SuffolkLITLab/docassemble-AssemblyLine-documentation/issues/134
[...]

PR #596, opened 2022-10-11
Use `user_role` over users == plaintiff

BryceStevenWilley:
When you are directly comparing the users to the plaintiffs, DA will try to gather both of the objects. This is fine if `users` is `plaintiffs`, but if the users are the defendants instead, it will trigger a infinite loop screen, since `other_parties` has been set to `plaintiffs`, and it's trying to gather `plaintiffs` before being able to fill `other_parties`.

Additionally, uses `user_role` instead of `user_ask_role`, which will always ask for the user's role in the case, even if it's a form that always done from one side, or we can detect what side of the case the user is on (as with e-filing).

Would like to do some additional testing with this on the dev server, but that might take a bit, since I don't know which interviews to test with. I'll make sure the ALAffidavit and the interpreter notice work correctly.

A minimal reproduction, with 57ee87e496109ba8f2c7c7175991cea38eb86100: Saying that you are a defendant will result in an error screen.

```yaml
include:
---
mandatory: True
code: |
  users.gather()
  other_parties.gather()
```

Results in the <summary>Following error:
<details>
<pre>
There appears to be an infinite loop. Variables in stack are x.first.
...
Needed definition of other_parties[0].name.first at 0.33673s
Tried to ask question at 0.33724s

(from ql_baseline.yml)

sets:
  - other_parties[i].name.first
  - other_parties[i].name.last
  - other_parties[i].name.middle
  - other_parties[i].name.suffix
id: names of opposing parties
question: |
  % if users==plaintiffs:
  Name of ${ ordinal(i) } **defendant** or respondent in this
  matter
  % else:
  Name of ${ ordinal(i) } **plaintiff** or petitioner in this matter
  % endif
fields:
  - code: |
      other_parties[i].name_fields(person_or_business='unsure')
</pre>
(repeated)
</details>
</summary>

Doesn't happen with this PR.

### DIFF
--- a/docassemble/AssemblyLine/data/questions/ql_baseline.yml
+++ b/docassemble/AssemblyLine/data/questions/ql_baseline.yml
@@ -148,10 +148,10 @@ code: |
     petitioners = other_parties
 ---
 depends on:
-  - user_role
+  - user_ask_role
 code: |
   # If the interview author didn't specify and it's a "court" or "other" case, we just ask
-  user_started_case = user_role == 'plaintiff'
+  user_started_case = user_ask_role == 'plaintiff'
 ---
 id: what is the user's detailed role in this case?
 comment: |
@@ -181,7 +181,7 @@ fields:
     datatype: radio
     choices:
       - Petitioner: petitioner
-      - Plaintiff: plaintiff      
+      - Plaintiff: plaintiff
       - I don't know: unknown
     show if:
       variable: user_detailed_role_started_case
@@ -804,7 +804,7 @@ fields:
 ---
 id: any other opposing parties
 question: |
-  % if user_role == 'plaintiff':
+  % if user_started_case:
   Is there any other **defendant** or respondent in this
   matter?
   % else:
@@ -821,7 +821,7 @@ sets:
   - other_parties[i].name.suffix
 id: names of opposing parties
 question: |
-  % if user_role == 'plaintiff':
+  % if user_started_case:
   Name of ${ ordinal(i) } **defendant** or respondent in this
   matter
   % else:
@@ -1503,12 +1503,12 @@ buttons:
 id: is the user the plaintiff or defendant?
 comment: |
   This question is asked if the interview can be used by both
-  a Plaintiff and a Defendant. It sets the value of `user_role`,
+  a Plaintiff and a Defendant. It sets the value of `user_role`, `user_started_case`,
   `plaintiffs`, `defendants`, `petitioners` and `respondents`.
   
   If you know your interview is exclusively
-  used by the Plaintiff or Defendant, set `user_role = 'plaintiff'` or 
-  `user_role='defendant'` in your code.
+  used by the Plaintiff or Defendant, set `user_ask_role = 'plaintiff'` or
+  `user_ask_role='defendant'` in your code.
 question: |
   Did you start this case, or are you responding to a case
   that someone else started?
--- a/docassemble/AssemblyLine/data/questions/ql_baseline.yml
+++ b/docassemble/AssemblyLine/data/questions/ql_baseline.yml
@@ -804,7 +804,7 @@ fields:
 ---
 id: any other opposing parties
 question: |
-  % if users==plaintiffs:
+  % if user_role == 'plaintiff':
   Is there any other **defendant** or respondent in this
   matter?
   % else:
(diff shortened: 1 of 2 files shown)


docassemble/AssemblyLine/data/questions/ql_baseline.yml
```diff
@@ -148,10 +148,10 @@ code: |
     petitioners = other_parties
 ---
 depends on:
-  - user_ask_role
```
nonprofittechy, line 151: `depends on` doesn't trigger the question, but it forces an update if the variable changes. i think we at least need to keep this trigger in place

nonprofittechy (commented):
Agreed that I want to see some test cases--this looks like it has a lot of subtle impacts to me.

docassemble/AssemblyLine/data/questions/ql_baseline.yml
BryceStevenWilley: The variable used in this change is now `user_role` though, so we'd want it to force an update if `user_role` changes, not if ` user_ask_role` changes. I'll make sure that the `depends on` is correct in the block that sets `user_role` though.

BryceStevenWilley:
Additional tests:

* ALAffidavitOfIndigency uses `list collect`, which this doesn't seem to affect, as the error isn't present with current AssemblyLine either. No change in behavior for this PR though
* With the interpreter notice, ran into this issue: https://github.com/SuffolkLITLab/docassemble-Interpreternotice/issues/19. Currently, you just can't be defendant, even though it asks you directly (because `user_ask_role` is encoded in the interview and it shouldn't). The bigger issue is that it's marked as `al_form_type = "starts_case"` which I don't think is the case. Will have to fix that additionally.

Any other ideas of interviews on AL that should be tested? @nonprofittechy   A lot of these problems are also related to https://github.com/SuffolkLITLab/docassemble-AssemblyLine/issues/181, just to link these two issues.

BryceStevenWilley:
Quick ping on this @nonprofittechy, is there any other interviews or anything else that you think should be tested?

nonprofittechy:
We need to figure out the impact of this change on existing forms that "correctly" directly trigger `user_ask_role`, and people who directly set `user_role` in the interview order block, as happens in the Weaver's default output.

BryceStevenWilley:
Started this a bit (found [one issue in the Weaver](https://github.com/SuffolkLITLab/docassemble-ALWeaver/issues/699)), but I'm not sure testing just Weaver stuff is good enough here, and all changes will likely break someone. 

Our own documentation is inconsistent on if you should directly set `user_role` or `user_ask_role`. The [definition on user_role](https://github.com/SuffolkLITLab/docassemble-AssemblyLine-documentation/blob/main/docs/framework/magic_variables.md#user_role) says you should directly set it, but ["Setting the form type"](https://github.com/SuffolkLITLab/docassemble-AssemblyLine-documentation/blob/973ac21bec11317f37a02c64295d739a5563519e/docs/customizing_interview.md#setting-the-form-type) says you should directly set `user_ask_role`. Additionally, I'm realizing that ILAO's appearance commonly uses `user_ask_role` in it's logic. 

So we can't rely on existing interviews to look like they come straight out of the weaver. Specifically, we can't rely on people having 1) gone through the weaver correctly, and 2) ever having fixed their code from the weaver if they went through incorrectly. If that's the case, this will break someone's interview, because `user_role` could be magically set from the Weaver, but never be used in the interview itself (it seems that none of AL actually uses the result of `user_role` right now, it would just show up in the output form likely). This PR would switch all their logic from depending on `user_ask_role` to `user_role`, which would break them.

To avoid all of this, I think I will just change the variable in this PR to be `user_started_case`, which is used in other logic. If someone is misusing `user_started_case` in their interview, their interview is currently broken. It doesn't solve for e-filing, but I guess I'll just have to deal with that case by case, avoiding `user_role`.

BryceStevenWilley:
Ready for another review. The only logic change now is from:
```mako
%if users == plaintiffs:
```
to

```mako
%if user_started_case:
```
Since `user_started_case` sets `plaintiffs = users` in the question library, this PR will make no additional semantic changes to existing interviews. I did add some comments discouraging the use of `user_role` in interviews, since we are using it less and less on the back end in favor of `user_ask_role` and `user_started_case`.

nonprofittechy (approved):

PR #601, opened 2022-10-31
Added black example hook + black opts to pyproject.toml

BryceStevenWilley:
Also moved mypy configs to the same pyproject.toml instead of a separate mypy.ini.

Don't know where to put the instructions, but all you need to do post cloning the repo is to copy the contents of `.github/hooks` into `.git/hooks`. There's not a way to allow people to clone hooks, you have to set them up manually.


nonprofittechy (commented):
Does this prevent uploading code that doesn't pass the `black` formatter test, or does it run `black` to fix the code?

nonprofittechy:
> > Does this prevent uploading code that doesn't pass the black formatter test, or does it run black to fix the code?
> 
> It prevents _committing_ code that doesn't pass the formatter. I don't think changing code in a pre-commit hook is standard practice. To fix it you should be able to just run `black .` though.

That would mean people who use the Playground would have to manually edit their code, right? I think that will be a tricky process to get adoption with, unfortunately.

BryceStevenWilley:
I think I misunderstood your previous question: this PR by itself doesn't do anything (besides setting config for black). Pre-commit hooks have to be manually set up on each clone, so this is like an "opt-in" (on the playground there's no way to opt-in, so things are unchanged there). You need to copy the contents of `.github/hooks` into `.git/hooks`.

I mostly made this PR because you were mentioning setting up pre-commit hooks, so I figured making sure we had the same hook would be a good start.

nonprofittechy:
> I think I misunderstood your previous question: this PR by itself doesn't do anything (besides setting config for black). Pre-commit hooks have to be manually set up on each clone, so this is like an "opt-in" (on the playground there's no way to opt-in, so things are unchanged there). You need to copy the contents of `.github/hooks` into `.git/hooks`.
> 
> I mostly made this PR because you were mentioning setting up pre-commit hooks, so I figured making sure we had the same hook would be a good start.

Perfect! Yes that is very helpful, thank you!

PR #602, opened 2022-10-31
Teams bump hook, in a github action

### LINKED ISSUE #547: Move the Teams announcement feature to GithubWorkflows
IMO the workflows have been working really well for us. With https://github.com/SuffolkLITLab/docassemble-AssemblyLine/commit/91d579d1a3c15283d3dd7557d61ce31478238780, we are now using it to publish to pypi when making github tags. The only feature that we don't get from the workflow yet is also making an announcement to teams. That should be able to be copied to the github workflow, using [this section of the bumpversion script](https://github.com/SuffolkLITLab/docassemble-AssemblyLine/blob/4c45d5cfe6be621898d765df61013d36c9661f7f/bumpversion.sh#L151-L217), which also has some benefits like being able to make the URL we push to a github secret.

BryceStevenWilley:
Will announce new tagged releases on pypi and github to teams. Fixes #547

A copy is working in my testing repo: https://github.com/BryceStevenWilley/docassemble-TestALPypi/actions/runs/3381057069/jobs/5614534406 (will be deleted after this is merged).

A screenshot of what the teams announcement will look like: 
Screenshot from 2022-11-02 16-55-56

The teams bumphook should already be set as an org wide secret.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,13 +1,17 @@
-name: Publish Python 🐍 distributions 📦 to PyPI and TestPyPI
+name: Publish Python 🐍 distributions 📦 to PyPI and TestPyPI and announce to teams
 
 on: push
 
 jobs:
   build-n-publish:
-    name: Build and publish Python 🐍 distributions 📦 to PyPI
     runs-on: ubuntu-20.04
+
+    env:
+      TEAMS_BUMP_WEBHOOK: ${{ secrets.TEAMS_BUMP_WEBHOOK }}
+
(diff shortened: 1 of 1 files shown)


nonprofittechy (approved):
Great!

BryceStevenWilley:
At some point I think I will factor out this workflow into a separate repo with just this action, so all of the other repos we want annoucements for (ALToolbox, ALWeaver, EFSPIntegration, FormFyxer, etc) can share this same workflow. But for now, I wanna make sure it works here for a few releases without any unanticipated behavior.

PR #604, opened 2022-11-02
Added a max size feature to the ExhibitDocument

BryceStevenWilley:
Also give suggested remaining upload size in human readable file sizes in errors and in between upload screens.

Test is in new file called `test_alexhibit_maxsize.yml`, as it's not easily tested in a unit test.

Made this because e-filing into Tyler will require size limits on the overall size of the submission, not just each individually uploaded file like we have now. Unfortunately, the existing approach of "send the documents as a downloadable link instead of sending in the email" won't work, as Tyler's system has a max limit, and I think we want to tread carefully about using external documents not in the filing, as it'll force the courts to act differently with our filings.

Also found a bug (that [I reported in the docassemble slack](https://docassemble.slack.com/archives/C7791ATUJ/p1667409800971949)) where `validation_error(...)` on show if'd file uploads don't work / show up at all. So I removed the validation errors that were currently doing that.

### DIFF
--- a/docassemble/AssemblyLine/data/questions/ql_baseline.yml
+++ b/docassemble/AssemblyLine/data/questions/ql_baseline.yml
@@ -16,7 +16,7 @@ objects:
   - attorneys: ALPeopleList
   - translators: ALPeopleList
   - debt_collectors: ALPeopleList
-  - creditors: ALPeopleList  
+  - creditors: ALPeopleList
   - spouses: ALPeopleList
   - parents: ALPeopleList
   - decedents: ALPeopleList.using(ask_number=True, target_number=1)
@@ -41,8 +41,8 @@ subquestion: |
   % endif
   
   Step 1. Answer questions that will fill in your form for you.  
-  Step 2. Preview the completed form.  
-  % if form_approved_for_email_filing:  
+  Step 2. Preview the completed form.
+  % if form_approved_for_email_filing:
   Step 3. Email the form to the court using this secure website and save copies
   for yourself for later reference.  
   % elif al_form_type in ['starts_case','existing_case','appeal']:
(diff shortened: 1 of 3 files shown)


docassemble/AssemblyLine/data/questions/ql_baseline.yml
```diff
@@ -1603,31 +1612,43 @@ id: exhibit i
...
     datatype: files
     maximum image size: 1024
-    image upload type: jpeg        
+    image upload type: jpeg
     accept: |
       "image/png, image/jpeg, .doc,.docx,application/msword,application/vnd.openxmlformats-officedocument.wordprocessingml.document,application/pdf,.pdf"   
 validation code: |
-  if sum(exhibit.size_in_bytes() for exhibit in x[i].pages) > (15 * 1024 * 1024):
+  this_doc_size = sum(a_page.size_in_bytes() for a_page in x[i].pages)
+  log(f"this doc: {this_doc_size}")
```
nonprofittechy, line 1627:
```suggestion
```

docassemble/AssemblyLine/data/questions/ql_baseline.yml
```diff
@@ -1672,6 +1701,11 @@ question: |
   You have ${ x.number_gathered() } document(s) so far. Do you have another document
   you want to upload?
 subquestion: |
+  % if hasattr(x, 'maximum_size'):
+  The total size of all exhibits must be less than ${ humanize.naturalsize(x.maximum_size) }.
+  You can upload ${ humanize.naturalsize(x.maximum_size - x.size_in_bytes())} more.
+  % endif
+
```
nonprofittechy, line 1705: I don't love this wording, but I don't have any better yet and it's better to warn in advance than do nothing.

nonprofittechy (approved):

docassemble/AssemblyLine/data/questions/ql_baseline.yml
BryceStevenWilley: Definitely agree that it could be better. I had to avoid "All exhibits must be less than ...", because that reads like each exhibit individually needs to be less than that.

PR #605, opened 2022-11-07
If address is empty, `on_one_line()` will be empty

BryceStevenWilley:
Originally, was ", ", which is def a bug.

Suggested by @mnewsted


mnewsted:
Thanks @BryceStevenWilley !

nonprofittechy (approved):

PR #606, opened 2022-11-07
No more implicit optionals in function signatures

BryceStevenWilley:
Apparently not allowed by PEP-484. Personally, I've gotten used to it, but I remember being confused by it, and the change isn't too bad.

Is the real fix to https://github.com/SuffolkLITLab/docassemble-AssemblyLine/actions/runs/3413494344/jobs/5680263993.

Open to leaving the option in, but I'm also always for more strict and direct types.


nonprofittechy (approved):
Looks good, happy to just follow the PEP here.

PR #607, opened 2022-11-08
don't cut off text for addendum in mid-word

### LINKED ISSUE #243: Use smarter character limits in overflow fields so words are not cut off mid-word
we can use the `textwrap` module. Here's a tiny mockup:

```
---
code: |
  import textwrap
---
mandatory: True
question: |
  ${ next(iter(textwrap.wrap("abc 123 4556 afs 12390",10, break_long_words=False))) }
```

This should not trigger for very small fields, but maybe when it is over 10 characters?

nonprofittechy:
Fix #243 

Sample interactive test:

```yaml
---
modules:
  - .al_document
---
mandatory: True
code: |
  text_input
  display_results
---
objects:
  - myfield: ALAddendumField.using(field_name="text_input")
---
question: |
  Enter some text
fields:
  - Overflow trigger: myfield.overflow_trigger
    datatype: integer
  - Text: text_input
    datatype: area
    rows: 5
---
event: display_results
question: |
  Results
subquestion: |
  
  ${ single_to_double_newlines(myfield.safe_value(overflow_message="...", preserve_newlines=True, preserve_words=True)) }
  
  ${ single_to_double_newlines(myfield.overflow_value(overflow_message="...", preserve_newlines=True, preserve_words=True)) }
  
```

### DIFF
--- a/docassemble/AssemblyLine/al_document.py
+++ b/docassemble/AssemblyLine/al_document.py
@@ -294,7 +294,7 @@ class ALAddendumField(DAObject):
                     return next(iter(retval)).rstrip() + overflow_message
                 return value.rstrip()[:max_chars] + overflow_message
             else:
-                return re.sub(r"\s+", " ", value).rstrip()
+                return value.rstrip()
 
         # If the overflow item is a list or DAList
         if isinstance(value, list) or isinstance(value, DAList):
--- a/docassemble/AssemblyLine/al_document.py
+++ b/docassemble/AssemblyLine/al_document.py
@@ -262,19 +262,13 @@ class ALAddendumField(DAObject):
         if isinstance(value, str):
             if len(value) > self.overflow_trigger:
                 if preserve_words:
-                    return (
-                        next(
-                            iter(
-                                wrap(
-                                    value,
-                                    width=max_chars,
-                                    replace_whitespace=True,
-                                    drop_whitespace=True,
-                                )
-                            )
-                        )
-                        + overflow_message
+                    retval = wrap(
+                        value,
+                        width=max_chars,
+                        replace_whitespace=True,
+                        drop_whitespace=True,
                     )
+                    return next(iter(retval)) + overflow_message
                 return (
                     re.sub(r"[\r\n]+|\r+|\n+", " ", value).rstrip()[:max_chars]
                     + overflow_message
--- a/docassemble/AssemblyLine/al_document.py
+++ b/docassemble/AssemblyLine/al_document.py
@@ -28,6 +28,7 @@ from docassemble.base.util import (
 )
 from docassemble.base.pdfa import pdf_to_pdfa
 from textwrap import wrap
+from math import floor
 
 __all__ = [
     "ALAddendumField",
@@ -147,6 +148,13 @@ class ALAddendumField(DAObject):
         """
         Try to return just the portion of the variable (list-like object or string)
         that is not contained in the safe_value().
+
+        Whitespace will be altered. If preserve_newlines is true, the return value may have newlines,
+        but double newlines and Windows style (\r\n) will be replaced with \n. Double spaces will replaced
+        with a single space.
+
+        If preserve_newlines is false, all whitespace, including newlines and tabs, will be replaced
+        with a single space.
         """
         # Handle a Boolean overflow first
         if isinstance(self.overflow_trigger, bool) and self.overflow_trigger:
@@ -161,11 +169,18 @@ class ALAddendumField(DAObject):
             _original_value=original_value,
             preserve_words=preserve_words,
         )
-        if isinstance(safe_text, str):
-            # Always get rid of double newlines, for consistency with safe_value.
-            value_to_process = re.sub(
-                r"[\r\n]+|\r+|\n+", r"\n", original_value
-            ).rstrip()
+        if isinstance(safe_text, str):               
+            max_lines = self.max_lines(input_width=input_width)
+
+            if preserve_newlines and max_lines > 1:
+                # we do our own substitution of whitespace, only double newlines and spaces
+                value_to_process = re.sub(
+                    r"[\r\n]+|\r+|\n+", r"\n", original_value
+                ).rstrip().replace("  ", " ")
+            else:
+                # textwrap.wrap(replace_whitespace=True) replaces all whitespace, not just double newlines and spaces
+                value_to_process = re.sub(r"\s+", " ", original_value).rstrip()
+
             if safe_text == value_to_process:  # no overflow
                 return ""
             # If this is a string, the safe value will include an overflow message. Delete
@@ -182,14 +197,11 @@ class ALAddendumField(DAObject):
         # Do not subtract length of overflow message if this is a list of objects instead of a string
         return original_value[self.overflow_trigger :]
 
-    def max_lines(self, input_width: int = 80, overflow_message_length=0) -> int:
+    def max_lines(self, input_width: int = 80) -> int:
         """
-        Estimate the number of rows in the field in the output document.
+        Calculate the number of lines of text that will fit in the specified input
         """
-        return (
-            int(max(self.overflow_trigger - overflow_message_length, 0) / input_width)
-            + 1
-        )
+        return floor(self.overflow_trigger / input_width)
 
     def value(self) -> Any:
         """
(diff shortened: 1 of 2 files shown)


docassemble/AssemblyLine/al_document.py
```diff
@@ -234,43 +238,43 @@ def safe_value(
...
+                        next(
+                            iter(
+                                wrap(
+                                    value,
+                                    width=max_chars,
+                                    replace_whitespace=True,
+                                    drop_whitespace=True,
+                                )
+                            )
+                        )
```
BryceStevenWilley, line 275: formatter decided to go a little crazy with this one 😅  Maybe break it into a few lines

docassemble/AssemblyLine/test_al_document.py
```diff
@@ -33,5 +33,29 @@ def test_upload_pdf(self):
...
+
+“You are welcome, all the same,” added this matron. “We have a very friendly and grateful feeling for your father, and it appears that we stand in great need of his counsel, which you are without doubt charged to communicate to us.”
+
+“I came from him,” I replied, “to talk over the affair with Madame d’Ionis.”        
+        """
+        define("fake_value", TEXT)
+        myfield = ALAddendumField(field_name="fake_value", overflow_trigger=120)
+        self.assertLessEqual(
+            len(myfield.safe_value(overflow_message=" [See addendum]")), 120
+        )
```
BryceStevenWilley, line 57: Good test, but I don't think it catches what this PR is trying to fix, that the words are unbroken?

docassemble/AssemblyLine/al_document.py
```diff
@@ -234,43 +238,43 @@ def safe_value(
...
-                else:
-                    return retval
+                # textwrap.wrap does all the hard work for us here
+                return " ".join(
+                    wrap(
+                        value,
+                        width=input_width,
+                        max_lines=max_lines,
+                        replace_whitespace=False,
+                        placeholder=overflow_message,
```
BryceStevenWilley, line 257: Not sure how this would interact with newlines in `value`, which people are likely to add in text areas I think.https://docs.python.org/3/library/textwrap.html#textwrap.TextWrapper.replace_whitespace says that it'll "cause strange output", which makes me a little hesitant.

BryceStevenWilley (commented):
IMO it needs a little more tests, I'm not sure I understand what `textwrap` does to multi-line input, which text areas are likely to get often

docassemble/AssemblyLine/test_al_document.py
nonprofittechy: Correct, I was thinking this was just a stub and it's not testing that yet. I can keep adding more tests though, probably worth doing while this is fresh

docassemble/AssemblyLine/al_document.py
nonprofittechy: It doesn't cause strange output in our usage, which ends up removing the line breaks after any get added.

We just use textwrap.wrap to figure out how many lines should fit, but we let the pdf renderer handle the actual soft line breaks in the output.

docassemble/AssemblyLine/al_document.py
```diff
@@ -234,43 +238,37 @@ def safe_value(
         )
         max_chars = max(self.overflow_trigger - len(overflow_message), 0)
 
-        # If there are at least 2 lines, we can ignore overflow trigger.
-        # each line will be at least input_width wide
+        # If we preserve newlines, we need to account for max_lines, not just max_chars
         if preserve_newlines and max_lines > 1:
             if isinstance(value, str):
+                max_lines = self.max_lines(
+                    input_width=input_width, overflow_message_length=0
```
BryceStevenWilley, line 245:
```suggestion
                    input_width=input_width, overflow_message_length=len(overflow_message)
```

docassemble/AssemblyLine/al_document.py
```diff
@@ -234,43 +238,37 @@ def safe_value(
...
         if isinstance(value, str):
             if len(value) > self.overflow_trigger:
+                if preserve_words:
+                    retval = wrap(
+                        value,
+                        width=max_chars,
+                        replace_whitespace=True,
+                        drop_whitespace=True,
+                    )
+                    return next(iter(retval)) + overflow_message
```
BryceStevenWilley, line 271:
```suggestion
                    return re.sub(r"\s+", " ", next(iter(retval))) + overflow_message
```

Found a bug: with overflow trigger of 23 and input text like this:

```
Testing here

with some very short words, but a whole lot of them, so it'll be over the overflow, over the flow yah know?
```
The safe value will be `
```
Testing here    with…
```
, but shows up as 
```
Testing here with...
```
because HTML collapses duplicate spaces automatically (which is why I had to use code blocks; with inline code, those two look the same!). The overflow message starts with `me very short words`, cuts off 2 characters.

I think this is because the overflow function handles new lines differently, and gets rid of the duplicate spaces, but `wrap`'s `replace_whitespace` and `drop_whitespace` doesn't.

docassemble/AssemblyLine/al_document.py
BryceStevenWilley: We aren't removing all of them, we're just collapsing things where multiple appear in a row. Test overflow of 90, and input:
```
At least 10% of people 

in this target audience do not know what this term means.
 
Therefore, you must avoid this term, explain this term, or link to an explanation.
```
Results in 3 lines of text, when there should only be 2, which could cause things to get cut off in the PDF.

docassemble/AssemblyLine/al_document.py
nonprofittechy: Good catch--I added a test case and made whitespace handling actually consistent between safe_ and overflow_value. HTML output won't actually matter, just the output in the DOCX and PDF, but that was also wrong.

nonprofittechy:
Tricky to test this given different way that whitespace (particularly newlines) happen in-browser input fields vs in a docstring in the unit test file, but I think this is all correct now! This now also fixes some minor, undiscovered but long-standing bugs in the addendum code too.

docassemble/AssemblyLine/al_document.py
```diff
@@ -290,9 +308,7 @@ def value_if_defined(self) -> Any:
         Return the value of the field if it is defined, otherwise return an empty string.
         Addendum should never trigger docassemble's variable gathering.
         """
-        if defined(self.field_name):
-            return value(self.field_name)
-        return ""
+        return showifdef(self.field_name, "")
```
nonprofittechy, line 311: @BryceStevenWilley please check my assumptions here--I think we had the wordier code to workaround an older, unnecessarily slow version of `showifdef()` but this simplification can be made now?

I'm looking at 

```python
def showifdef(var: str, alternative=''):
    """Returns the variable indicated by the variable name if it is
     defined, but otherwise returns empty text, or other alternative text.
     """
    # A combination of the preambles of defined and value
    str(var)
    if not isinstance(var, str):
        raise Exception("showifdef() must be given a string")
    if not re.search(r'[A-Za-z][A-Za-z0-9\_]*', var):
        raise Exception("showifdef() must be given a valid Python variable name")
    try:
        return eval(var, {})
    except:
        pass
    if re.search(r'[\(\)\n\r]|lambda:|lambda ', var):
        raise Exception("showifdef() is invalid: " + repr(var))
    return _defined_internal(var, DefCaller.SHOWIFDEF, alt=alternative)
```

and this seems to be a safe change but please double-check

docassemble/AssemblyLine/al_document.py
```diff
@@ -229,54 +258,43 @@ def safe_value(
...
+                        value,
+                        width=max_chars,
+                        replace_whitespace=True,
+                        drop_whitespace=True,
+                    )
+                    return next(iter(retval)).rstrip() + overflow_message
+                return value.rstrip()[:max_chars] + overflow_message
             else:
-                return re.sub(r"[\r\n]+|\r+|\n+", " ", value).rstrip()[:max_chars]
+                return re.sub(r"\s+", " ", value).rstrip()
```
nonprofittechy, line 297:
```suggestion
                return value.rstrip()
```

docassemble/AssemblyLine/al_document.py
BryceStevenWilley: I'm not sure if we were working around anything, but this literally used to be the code of [showifdef](https://github.com/jhpyle/docassemble/blob/595ad0a08572c0188f9ab49b1a6af99a3621cd42/docassemble_base/docassemble/base/functions.py#L4443-L4445), so switching would make things slightly faster. Haven't come across any bugs in the new implementation yet, and it's bin in there for ~a year or so I think, so :+1: from me.

BryceStevenWilley (approved):
LGTM. The only weird thing was that `preserve_newlines` isn't really respected if `max_lines` < 1, but that makes sense IMO, maybe worth documenting, but up to you.

nonprofittechy:
> LGTM. The only weird thing was that `preserve_newlines` isn't really respected if `max_lines` < 1, but that makes sense IMO, maybe worth documenting, but up to you.

I'm going to add some better documentation on the public documentation site

PR #613, opened 2022-11-17
Change OCR to use ocrmypdf in the background

BryceStevenWilley:
Triggers an event in `al_document.yml`, which calls back into a module.

`--skip-text` is the fastest and probably what we want. Tried with `--redo-ocr` , but it was very slow (4 minutes for just 14 pages), and only 3 of those pages had any text. Don't think `--force-ocr` is what we want.

Difficult to get performance on it, since it runs in celery on a random worker, so I can't really use `py-spy`, but it's probably good enough for now.o

Still leaving things flexible as to when in an interview you want to wait for the fully ocr'd text; at these speeds, I'd be okay with making people wait before the final screen, but I'll also try putting e-filing in the background, if it's not too hard.

I've been testing with:

```yml
---
include:
---
objects:
  - exhibit_doc: ALExhibitDocument.using(title="Exhibits", filename="exhibits", bates_prefix="EX-", include_exhibit_cover_page=True, include_table_of_contents=True, add_page_numbers=True, auto_ocr=True, enabled=True)
---
mandatory: True
code: |
  exhibit_doc.exhibits.has_exhibits
  exhibit_doc.exhibits.gather()
  start
  if not exhibit_doc.ocr_ready():
    waiting_screen
  end = current_datetime()
  all_done
---
event: waiting_screen
reload: True
question: |
  <i class="fas fa-cog fa-spin"></i> Please wait while we process your documents . . . <i class="fas fa-cog fa-spin"></i>
---
event: all_done
question: Done!
subquestion: |
  ${ exhibit_doc.as_pdf() }
  
  ${ end - start }
---
code: |
  start = current_datetime()
---
```

### DIFF
--- a/docassemble/AssemblyLine/al_document.py
+++ b/docassemble/AssemblyLine/al_document.py
@@ -1511,17 +1511,24 @@ class ALExhibit(DAObject):
 
     def _start_ocr(self):
         """
-        Starts the OCR (optical character resolution) process on the uploaded documents.
+        Starts the OCR (optical character recognition) process on the uploaded documents.
+        This adds a searchable text layer to any images of text that have been uploaded.
 
         Makes a background action for each page in the document.
         """
         if len(self.pages):
-            # We cannot OCR in place. It is too fragile.
             self.ocr_version = DAFile(self.attr_name("ocr_version"))
             self.ocr_version.initialize(filename="tmp_ocrd.pdf")
-            self.ocr_status = background_action(
-                "al_exhibit_ocr_pages", to_pdf=self.ocr_version, from_file=self.pages
-            )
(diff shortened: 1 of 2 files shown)


nonprofittechy:
LGTM, but I'm going to put it behind a configuration option--don't want to make it more complicated for folks to install AssemblyLine, and I think none of our partners currently are using OCR of any kind

BryceStevenWilley:
That works, it slipped my mind to just have a fallback if we detect that ocrmypdf doesn't there (`ocrmypdf --version`), but config is probably better.

BryceStevenWilley:
Tested with and without config, is good now. Another small thing that's slightly better DA's OCR is the fact that images stay in color and don't get converted to grayscale.

PR #617, opened 2022-11-21
Add visible trigger variable to the developer-only heads up display

### LINKED ISSUE #616: Add trigger variable to the developer info HUD
See also #232

nonprofittechy:
Fix #616

### DIFF
--- a/docassemble/AssemblyLine/data/questions/al_visual.yml
+++ b/docassemble/AssemblyLine/data/questions/al_visual.yml
@@ -43,6 +43,7 @@ default screen parts:
   pre: |
     % if get_config('debug'):
     `id: ${ user_info().question_id }`  
+    `Variable triggering this screen: ${ user_info().variable }`  
     `Package: ${ user_info().package } ${ package_version_number }; ${ al_version }`
     <div data-variable="${ encode_name(str( user_info().variable )) }" id="sought_variable" aria-hidden="true" style="display: none;"></div>
     % endif


plocket (approved):
Just noting that in the code below, we used the `str()` function. I'm not sure why we took that precaution, so I don't know if it's necessary. Otherwise, it looks good!

docassemble/AssemblyLine/data/questions/al_visual.yml
```diff
@@ -43,6 +43,7 @@ default screen parts:
   pre: |
     % if get_config('debug'):
     `id: ${ user_info().question_id }`  
+    `Variable triggering this screen: ${ user_info().variable }`
```
BryceStevenWilley, line 46: We'll want to do maybe a bit more than just `user_info().variable` here, since that will just show `x.my_attr` for generic questions.

Trying to reconstruct the actual variable inline can get confusing, but this is a start:

```suggestion
    `Variable triggering this screen: ${ user_info().variable } ${("x = " + str( x.instanceName )) if defined('x') else ''} ${ ("i =" + str( i )) if defined('i') else ''}`  
```

https://github.com/SuffolkLITLab/ALKiln/issues/582 has some more details on a similar thing in Kiln.

nonprofittechy:
> Just noting that in the code below, we used the `str()` function. I'm not sure why we took that precaution, so I don't know if it's necessary. Otherwise, it looks good!

I think that's for purposes of `encode_name` - the conversion to `str` is implicit with plain mako tags.

docassemble/AssemblyLine/data/questions/al_visual.yml
nonprofittechy: Not sure if this actually adds something in this context -- can you give an example of a screen where it's useful so I can better evaluate?

image

BryceStevenWilley: shoot, never mind, I was wrong. I think I was thinking of some of the kiln issues, which are getting the variable names from the fields; those would look more like `x[i].name.first`. I thought `user_info.variable()` had the same issue but it doesn't, sorry.

PR #619, opened 2022-11-22
First version of DACatchAll

### LINKED ISSUE #618: Add a series of DACatchall placeholder question
Today at deep dive we discussed ways to make this feature a teaching tool for new developers

nonprofittechy:
For now, this adds an optional include: `ql_catchall.yml` which a developer can incorporate to turn on the DACatchAll feature. I think it might help current students of myself and @colarusso if we integrate this with the MassAccess theme and install on the dev server, but we should try it out for a bit before deciding if it's something to enable globally.

I'm especially interested to see what we'll want to add to handle things like: invalid object references (e.g., we can probably turn unknown object references into ALIndividuals and unknown list references into ALPeopleLists). But I'm not sure best how to handle that yet--maybe we'll come up with some regexes when we figure out what common developer errors are.

Fix #618

### DIFF
--- a/docassemble/AssemblyLine/data/questions/ql_catchall.yml
+++ b/docassemble/AssemblyLine/data/questions/ql_catchall.yml
@@ -1,0 +1,137 @@
+---
+features:
+  use catchall: True
+---
+###################### Catchall questions ###########################################
+---
+generic object: DACatchAll
+question: |
+  ${ x.object_name() }?
+fields:
+  - ${ x.object_name() }: x.value
+subquestion: |
+  % if get_config("debug"):
+  ${ collapse_template(x.catchall_question_explanation) }
+  % endif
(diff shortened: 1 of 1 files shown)


BryceStevenWilley:
Am I using it right?  I tried this code, but it just skips right to the end, without asking me any questions.

```yaml
include:
  - assembly_line.yml
  - ql_catchall.yml
---
mandatory: True
code: |
  new_var
  new_per.name.first
  users[0].job_title
```

BryceStevenWilley:
> Weird, but it looks like you need to reference the variable in a context that displays on the screen

That's fair. It doesn't look like there's any bugs with it picking up a variable in a code block either.It just doesn't want to actually gather it until it's used, not just defined. Weird, but doesn't break interview logic.

Issue #232, opened 2021-07-02
Meta: safety


nonprofittechy:
We should think about various ways to prevent users from ever seeing an error screen. We should also identify all of the ways that docassemble interviews can be unsafe.

## Possible errors

* Missing variable that isn't defined in the YAML file
* key or item index that doesn't exist
* Infinite loop
* Screen that is too slow (mostly with files: document assembly, upload, and concatenation/OCR of files)
* API query that fails (e.g., the `address autocomplete` feature breaks in a way that prevents filling in addresses)
* Stuck on screen that cannot provide answer (`show if` bug)
* Unsafe handling of datatypes - string concatenation to a None is common. `None` handling in general.

## Possible solutions

1. Create a style guide that discourages the use of indexes and keys in brackets in favor of using `.item(n)`, `next(iter(some_list))` and `.get('key')`. 
2. Create some DACatchAll objects to help resolve undefined variable names. https://docassemble.org/docs/initial.html#use%20catchall
3. Create the linter we have previously discussed to try to resolve some `show if` bugs.
4. Aggressively catch JavaScript errors that could prevent submission.
5. Continue our work to set standards around testing.
6. Create a custom error action that always allows the user to download the form in-progress, with variable set to empty (https://docassemble.org/docs/initial.html#metadata - search for error action)
7. Use more typing and figure out if it can help identify potential errors, especially in Mako where existing code probably won't be enough.
8. Use background actions for slow tasks.
9. Increase timeouts in nginx.
10. Replace the 504 gateway timeout page with one that has "nicer" branding, a retry button, and includes contact information to get help. (started here: https://github.com/SuffolkLITLab/docassemble-ALToolbox/blob/main/docassemble/ALToolbox/data/static/custom_504.html)
10. Prevent uploading large files. (https://github.com/SuffolkLITLab/docassemble-AssemblyLine/pull/332)
11. Add "skip undefined" option for DOCX templates: (https://github.com/jhpyle/docassemble/pull/464)

Consider this issue a place to brainstorm--we can create more specific issues after we do this work.

nonprofittechy:
https://github.com/SuffolkLITLab/docassemble-AssemblyLine/pull/332

nonprofittechy:
Do we want to add a "Download form in progress" link to the menu?

nonprofittechy:
We've implemented many of these ideas and I think the issue itself is now stale. Fine for us to revisit and mine new ideas later though.

PR #622, opened 2022-11-22
Add a custom error action

### LINKED ISSUE #615: Add an "exit" link to the main menu
We have a "Start over" option, but we should also have an "exit" option for people who are not logged in but want to delete their information when they finish using an app.

Exit might not be the right word; maybe "delete my answers"

### LINKED ISSUE #620: Create a custom error action
See #232 

Basic features:
1. Show the download screen, with
    * An extra “key” in the ALDocument that’s the fallback version
    * Do the download generation in the background, but show some kind of waiting screen
1. What information would people want on this error screen?
    * How to contact us
    * How to contact the court?
         * Implies a new variable to track the court’s info
     * Ideally, the original form link if it was a paper PDF form
     * Ideally, the document with as much filled in as possible
1. Needs some visual style that makes it clear the user ran into an error
1. ~Will this prevent the built-in feature to email an error report in the config?~
    * Should be something we can do manually, but we need to check
    * Update: Just looked in server.py and the error email is independent of the custom error action.

nonprofittechy:
Creates a custom error page, with an optional link to download the form in progress. The download depends on the `skip undefined` parameter being correctly used on the attachment block.

In addition, this adds an "Exit and delete answers" link (#615) and a link to download the forms in-progress at any time, per discussion on Coding Deep Dive on 11/20/22 and broader discussion in #232

Fix #620 
Fix #615 

image
image

`DACatchAll` is turned on on our development server. If you want to simulate an error, one that I have been trying is triggering a hashing error. E.g., `users[{}]` in the interview order block.

I probably want to keep thinking about this a bit more over the next week or two, but it should be ready to look at. Specifically, I might want to add some features to the GithubFeedbackForm

### DIFF
--- a/docassemble/AssemblyLine/data/questions/al_visual.yml
+++ b/docassemble/AssemblyLine/data/questions/al_visual.yml
@@ -73,7 +73,7 @@ code: |
   if _internal.get('steps') > 1:
     menu_items = al_menu_items
   else:
-    menu_items = al_menu_items[1:] # Skip the "start over" menu item
+    menu_items = al_menu_items[3:] # Skip the "start over", download, and exit and erase menu item
 ---
 reconsider: True
 code: |
@@ -81,6 +81,12 @@ code: |
     {"url": url_ask(['al_start_over_confirmation','al_start_over']),
     "label": "Start over"
     },
+    {
+      "url": url_ask(['al_exit_logout_confirmation', 'al_exit_logout']),
+      "label": "Exit and delete my answers"
+    },
+    {"url": url_action('al_error_action_download_screen'), 
+    "label": "Download current progress"},
   ]
   
   # Answer set feature is behind a server-wide configuration option
@@ -102,6 +108,10 @@ code: |
     
   al_menu_items = al_menu_items_tmp
   del al_menu_items_tmp
+---
+event: al_exit_logout
+code: |
+  command("exit_logout")
 ---
 event: al_start_over
 code: |
@@ -349,6 +359,22 @@ continue button label: |
   Start over
 continue button field: al_start_over_confirmation
 ---
+id: exit and erase answers
+question: |
+  Do you want to delete your answers?
+subquestion: |
+  Your progress on this document will be completely erased. Any
+  actions that were scheduled, such as reminders from this website,
+  will also be removed.
+  
+  Click "Exit and delete my answers" to delete your progress,
+  or "Back" if you want to keep working.
+back button label: |
+  Back
+continue button label: |
+  Exit and delete my answers
+continue button field: al_exit_logout_confirmation
+---
 ################################### Custom error action ################################
 ---
 metadata:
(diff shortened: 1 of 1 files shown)


docassemble/AssemblyLine/data/questions/al_visual.yml
```diff
@@ -73,15 +73,29 @@ code: |
   if _internal.get('steps') > 1:
     menu_items = al_menu_items
   else:
-    menu_items = al_menu_items[1:] # Skip the "start over" menu item
+    # Skip the "start over", download, and exit and erase menu item
+    if get_config("assembly line", {}).get("enable incomplete downloads"):
+      menu_items = al_menu_items[3:]
+    else:
+      menu_items = al_menu_items[2:]
```
BryceStevenWilley, line 80: Do we want this in the config, or do we want it per-interview? I can easily imagine someone wanting it to apply to some interviews but not others.

nonprofittechy: That's fair, I can make it two level

docassemble/AssemblyLine/data/questions/al_visual.yml
```diff
@@ -348,7 +366,119 @@ back button label: |
...
+  % if al_enable_incomplete_downloads:
+  * Try to [download your work in progress](${ url_action("al_error_action_download_screen") })
+  % endif
+  % if MAIN_METADATA.get("original_form") and not MAIN_METADATA.get("original_form").strip() == "None":
+  * [Download a blank "${ all_variables(special='metadata').get("title").strip() }"](${ MAIN_METADATA.get("original_form") } )
+  % endif
+  * Visit the [${ AL_ORGANIZATION_TITLE } home page](${ AL_ORGANIZATION_HOMEPAGE }) and try another interview
+  % if al_show_email_to_user_on_errors and get_config("error notification email"):
+  * [Send us an email](mailto:${ get_config("error notification email") }?subject=User%20error%20report%3A%20${ all_variables(special='metadata').get("title") }%20${ user_info().filename }) (Note: you may not get a reply)
+  % endif
```
BryceStevenWilley, line 430: The chance that people actually email us is fairly low, and the auto-emails from errors in interviews have been really useful. Anyway we can just auto-send an email if someone sees this screen, with the same debugging information they would see, or would that require an upstream change

BryceStevenWilley (commented):
Haven't had time to really play with it, but it LGTM. I do think that the error emails are really important for production interviews; I don't think this should go to prod until we figure that out. If we want to try this on the dev server for a bit that would be fine, but we might forget that we don't want it on prod if a hotfix comes in, so maybe keep this open until we can figure out the errors emails?

docassemble/AssemblyLine/data/questions/al_visual.yml
nonprofittechy: Afaict the error emails still happen when you have a custom error action. Just from looking at the code the emails happen before the error action gets triggered

nonprofittechy: Just gave this a test by turning on notifications on the dev server and I can confirm it works even with the custom error action

PR #634, opened 2022-12-06
Fix unittest action

BryceStevenWilley:
Also bump to python 3.9: closer to still working with python
3.8 for older docassemble versions, but still on by default
in ubuntu 22.04, the new test runners on github.

Should also try to cache pip dependencies. Should work,
will need to re-run the job to see if that works, (though it's
less of a priority).


BryceStevenWilley:
Should subsume #633.

PR #639, opened 2022-12-12
Don't force gathering users[0].address to gather users[1].address or users[0].mailing_address

### LINKED ISSUE #636: Don't always ask for user[0].address before other users
Something that Matt brought up in the appearance interview: because the `user[i].address` gives [a "Same as your address" option](https://github.com/SuffolkLITLab/docassemble-AssemblyLine/blob/06d5a9d774cf35b8eaf3083f8206688ece422eda/docassemble/AssemblyLine/data/questions/ql_baseline.yml#L480), it will always ask for `user[0].address` before everyone elses, even when that's out of order.

Since we might not need the filler's address on the form, asking the question gets more info than necessary. We should be able to fix it by adding `if: defined("user[0].address.address")` to the current question and adding another question without the "Same as your address" option. The fix does assume that authors want more exact control over what gets gathered in their interview instead of helping them along.

Just wanted to get some confirmation that we'd be okay with this before making a PR.

nonprofittechy:
Fix #636

### DIFF
--- a/docassemble/AssemblyLine/data/questions/ql_baseline.yml
+++ b/docassemble/AssemblyLine/data/questions/ql_baseline.yml
@@ -452,12 +452,15 @@ fields:
   - My mailing address is: users[0].mailing_address
     datatype: object_radio
     choices:
-      - users[0].address
+      - users[0].address if defined("users[0].address.address") else None
     object labeler: |
       lambda y: y.on_one_line()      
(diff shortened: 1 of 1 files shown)


BryceStevenWilley (approved):
LGTM. Unfortunate that we need `defined("users[0].address.address")` in there multiple times, but oh well.

Issue #581, opened 2022-09-22
Add a question for ALPeopleList[i].attorney


nonprofittechy:
It should prompt for bar number and office name, and it should allow a user to load information from a stored profile when we implement that feature eventually.

nonprofittechy:
It may make sense to have two separate screens-one for name and organization name, and one for the other details.

nonprofittechy:
Actually should consider making this plural - `.attorneys`

nonprofittechy:
I think `attorneys` is the right attribute name to make.

PR #640, opened 2022-12-13
Add simple service of process questions

### LINKED ISSUE #413: Add a "service_address" attribute to an ALIndividual

### LINKED ISSUE #597: Create a question for serving an opposing party
The purpose would be to help complete a certificate of service. It should let someone specify if they are going to serve the opposing party in hand, by email, etc., as well as whether they are serving in person or by their attorney.

It should be possible to specify the service method for each opposing party.

nonprofittechy:
Fix #597, fix #413 

At some point, this should link up with #581 (although note that  https://github.com/IllinoisLegalAidOnline/docassemble-Appearance/blob/main/docassemble/Appearance/data/questions/appearance.yml uses the word "lawyer" instead of "attorney" which might be better.

### DIFF
--- a/docassemble/AssemblyLine/al_general.py
+++ b/docassemble/AssemblyLine/al_general.py
@@ -459,6 +459,7 @@ class ALIndividual(Individual):
     previous_addresses: ALAddressList
     other_addresses: ALAddressList
     mailing_address: ALAddress
+    service_address: ALAddress
 
     def init(self, *pargs, **kwargs):
         super(ALIndividual, self).init(*pargs, **kwargs)
@@ -474,6 +475,8 @@ class ALIndividual(Individual):
             self.initializeAttribute("other_addresses", ALAddressList)
         if not hasattr(self, "mailing_address"):
             self.initializeAttribute("mailing_address", ALAddress)
+        if not hasattr(self, "service_address"):
+            self.initializeAttribute("service_address", ALAddress)
 
     def signature_if_final(self, i: str) -> Union[DAFile, str]:
         if i == "final":
--- a/docassemble/AssemblyLine/data/questions/ql_baseline.yml
+++ b/docassemble/AssemblyLine/data/questions/ql_baseline.yml
@@ -29,6 +29,10 @@ objects:
   - trial_court_address: Address
   - appeals_court_address: Address
 ---
+generic object: ALIndividual
+objects:
+  - x.service_address: ALAddress
+---
 id: basic questions intro screen
 decoration: form-lineal
 question: |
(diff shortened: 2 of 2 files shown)


docassemble/AssemblyLine/data/questions/ql_baseline.yml
```diff
@@ -536,6 +536,81 @@ fields:
...
+  - ${ x }'s service address is: x.service_address
+    datatype: object_radio
+    choices:
+      - x.address if defined(x.address.attr_name("address")) else None
+      - x.mailing_address if defined(x.mailing_address.attr_name("address")) else None
+    object labeler: |
+      lambda y: y.on_one_line()      
+    none of the above: |
+      Somewhere else
+    disable others: True
```
BryceStevenWilley, line 553: Is the show if on address and mailing address you used in #639, i.e.
```yml
   show if:
      code: |
        defined("x.address.address") and defined("x.mailing_address.address")
```
not needed here?

nonprofittechy: No, if the only option is `None` Docassemble won't display the field.

docassemble/AssemblyLine/data/questions/ql_baseline.yml
```diff
@@ -536,6 +536,81 @@ fields:
...
+data:
+  e-filing system: 
+    question: E-filing system
+    document: e-filing system
+  mail:
+    question: Mail
+    document: mail
+  hand:
+    question: In-hand delivery
+    document: in-hand
```
BryceStevenWilley, line 592: Is "in-hand" the right way of saying this? Motion to stay eviction just uses "hand-delivery", both on the document and in the question, it seems slightly easier to read

docassemble/AssemblyLine/al_general.py
```diff
@@ -474,6 +475,8 @@ def init(self, *pargs, **kwargs):
             self.initializeAttribute("other_addresses", ALAddressList)
         if not hasattr(self, "mailing_address"):
             self.initializeAttribute("mailing_address", ALAddress)
+        if not hasattr(self, "service_address"):
+            self.initializeAttribute("service_address", ALAddress)
```
BryceStevenWilley, line 479: I think there might also need to be an object block in the yml file, just
```yml
objects:
  - x.service_address: ALAddress
```

I've been bitten a few times when not putting stuff like that there, and breaking existing sessions of interviews, even if that's not likely to happen in this case

BryceStevenWilley (approved):
LGTM, not sure whether to try to encapulate the logic for authors a bit more, with a block like `x.service_info_gathered`, which would make sure that `x.email` exists if they selected email service, would automatically ask for the address of mailing service is used, etc. But maybe that's too complicated for different jurisdictions (I know IL needs time of day for hand delivery and stuff like that)

PR #642, opened 2022-12-13
Remove 'name' from label for name - Fix #589

### LINKED ISSUE #589: Change labels of the name questions from "First name", "Middle name" etc to "First"?
https://github.com/nonprofittechy/docassemble-HousingCodeChecklist/issues/237

In a user test session, user typed their full name into the "First name" field. Possibly because they saw "Name" closest to the label.

nonprofittechy:
Fix #589

### DIFF
--- a/docassemble/AssemblyLine/data/questions/ql_baseline.yml
+++ b/docassemble/AssemblyLine/data/questions/ql_baseline.yml
@@ -237,7 +237,7 @@ code: |
 generic object: ALIndividual
 template: x.first_name_label
 content: |
-  First name
+  First
 ---
 generic object: ALIndividual
 template: x.middle_name_label
@@ -247,7 +247,7 @@ content: |
 generic object: ALIndividual
 template: x.last_name_label
 content: |
-  Last name
+  Last
 ---
 generic object: ALIndividual
 template: x.suffix_label


BryceStevenWilley:
TBH I'm still hesitant about this change; I feel like we should be giving more information, not taking it away in cases like this. Haven't made an issue yet, but I'm still in favor of ["First or Given name"](https://designsystem.digital.gov/patterns/create-a-user-profile/name/#pattern-preview), from the USWDS.

Additionally, [one of our questions currently](https://github.com/SuffolkLITLab/docassemble-AssemblyLine/blob/a10aaa7e66bf57ea1db3a693ad4857f7047b88ed/docassemble/AssemblyLine/data/questions/ql_baseline.yml#L411), uses wording that doesn't make sense with just `First` ("Who is the Nth person in your case?").

If I remember correctly, #589 was an error from a single person; I think we should either see if it pops up multiple times with other people before merging.

nonprofittechy:
> TBH I'm still hesitant about this change; I feel like we should be giving more information, not taking it away in cases like this. Haven't made an issue yet, but I'm still in favor of ["First or Given name"](https://designsystem.digital.gov/patterns/create-a-user-profile/name/#pattern-preview), from the USWDS.
> 
> Additionally, some of our questions currently use wording that doesn't make sense with just `First` (["Who is the Nth person in your case?"](https://github.com/SuffolkLITLab/docassemble-AssemblyLine/blob/a10aaa7e66bf57ea1db3a693ad4857f7047b88ed/docassemble/AssemblyLine/data/questions/ql_baseline.yml#L411), and the "Person or Business" question in particular lose a lot of context without "name").
> 
> If I remember correctly, #589 was an error from a single person; I think we should either see if it pops up multiple times with other people before merging.

Hopefully will do some more user testing on this in the new year. But it was a compelling result--enough that I'm making the switch in that particular interview already.

BryceStevenWilley:
> But it was a compelling result--enough that I'm making the switch in that particular interview already.

I think I am a little confused; I know the issue what that someone filled in their full name in the first text field, but if I'm remembering correctly, you didn't directly ask them why they did that, just that it might have been the fact that "name" is the closer word to the field, right? Am I misremembering that?

Issue #552, opened 2022-08-07
Use the  `datatype: al_international_phone` for phone numbers as the default input type


nonprofittechy:
I've been using this a couple months and it's mature enough for use. The only limit is that it doesn't work with the `default:` modifier, but we don't use a `default` phone number in any of the question library questions.

BryceStevenWilley:
It's not in yet, but I had been planning on attempting to use `default` for phone numbers (and emails) with people who are logged in with Tyler E-filing system. I guess I could always overwrite the question library just for the efiling interviews but just a though.

nonprofittechy:
This would also interact with #418.

Still worth doing though, I think, as the international phone input is better in most ways than the current default.

PR #643, opened 2022-12-13
Create a framework to simplify adding sending users reminders and other alerts

### LINKED ISSUE #333: Create some wrapper code to help developers remind users of important dates to return/follow-up on next steps
https://github.com/SuffolkLITLab/EfileProxyServer/issues/1

Create reminders

Use cases:

A legal process has a beginning, middle and end but the Docassemble interview might only help someone with one part of the process. Reminders can help users complete the full legal process even if the interview itself can't complete every step.

1. Come back and file your form
2. Did the opposing party (or the court/forum) do the thing they were supposed to do?
    - Did you get the paperwork you asked for
3. You are about to hit a deadline for some action triggered by what you did in our interview. Take your next step
4. You started the interview and got to a screen asking you to come back later--come back later!

Could think of this mostly as a series of "next step" links to other interviews on the server.

nonprofittechy:
Fix #333

Opening as a draft for now, just to let other folks play around with the structure and see if the API is reasonable.

Here's a small interview I was using to experiment with launching the reminders manually:

You may want to try copying the format of the `al_reminders` block to add some of your own reminders, add additional reminder templates, etc.

```yaml
---
include:
  - assembly_line.yml
  - al_reminders.yml
---
mandatory: True
code: |
  users[0].mobile_number
  al_user_wants_reminders
---
variable name: al_reminders
data:
  filing_documents:
    description: Filing these documents
    date: ${ today().minus(days=3).format("yyyy-MM-dd") }
    email template: al_reminder_filing_template
    sms template: al_reminder_filing_template  
---
mandatory: True
question: |
  Click the button to launch the event
subquestion: |
  ${ action_button_html(url_action("cron_daily"), label="Launch cron") }
```

Edit: this is the API. Edit the block `al_reminders` and add a dictionary like this:

```yaml
variable name: al_reminders
data:
  filing_documents:
    description: Filing these documents
    date: ${ today().plus(days=3).format("yyyy-MM-dd") }
    email template: al_reminder_filing_template
    sms template: al_reminder_filing_template
```

All the keys are treated as strings. `email template` and `sms template` should refer to the name of a `template` block in the YAML, with a `subject` and `content` key. `date` should be a calculation that results in an ISO formatted date. This is just expected to be a dictionary, so you can edit and create it at any point in the interview. Note though that the date calculation gets made in-place here, implying that you've gathered all of the facts needed to provide a specific date.

### DIFF
--- a/docassemble/AssemblyLine/data/questions/al_reminders.yml
+++ b/docassemble/AssemblyLine/data/questions/al_reminders.yml
@@ -54,7 +54,7 @@ subquestion: |
   We can send you a reminder, by text or email,
   about ${ next(iter(al_reminders.values()))["description"][:1].lower() }${ fix_punctuation(next(iter(al_reminders.values()))["description"][1:]) }
   % else:
-  about:
+  We can send you a reminder, by text or email, about:
 
   % for reminder in al_reminders.values():
   * ${ reminder["description"]}
@@ -108,17 +108,17 @@ code: |
 ---
 template: al_reminder_initial_sms_template
 content: |
-  You asked to get a reminder of important dates when you used "${ all_variables(special='titles').get('title', AL_ORGANIZATION_TITLE) }".
+  You asked to get a reminder of important dates when you used "${ all_variables(special='metadata').get('title', AL_ORGANIZATION_TITLE) }".
   You will get the reminders at this phone number. 
   
-  [Unsubscribe](${ interview_url_action('al_reminders_unsubscribe_emails') }) to stop reminders from this
+  Click to unsubscribe ${ interview_url_action('al_reminders_unsubscribe_emails') } to stop reminders from this
   interview or reply STOP if you no longer want any reminders from ${ AL_ORGANIZATION_TITLE }.
 ---
 template: al_reminder_initial_email_template
 subject: |
   You are now signed up for reminders
 content: |
-  You asked to get a reminder of important dates when you used "${ all_variables(special='titles').get('title', AL_ORGANIZATION_TITLE) }".
+  You asked to get a reminder of important dates when you used "${ all_variables(special='metadata').get('title', AL_ORGANIZATION_TITLE) }".
   You'll get the reminders using this email address.
 
   You will get about ${ len(al_reminders) + 1 } emails in total (including this one).
@@ -133,9 +133,12 @@ code: |
 id: unsubscribe success
 event: al_reminders_display_unsubscribe_success
 question: |
-  You are now unsubscribed from reminders from "${ all_variables(special='titles').get('title', AL_ORGANIZATION_TITLE) }"
+  You are now unsubscribed from reminders from "${ all_variables(special='metadata').get('title', AL_ORGANIZATION_TITLE) }"
 subquestion: |
   You will no longer get any reminders of important dates by email or text message about this interview.
+
+  [Re-subscribe](${ url_action('al_reminders_resubscribe_emails') })
+back button: False  
 ---
 event: al_reminders_resubscribe_emails
 code: |
@@ -145,14 +148,15 @@ code: |
 id: resubscribe success
 event: al_reminders_display_resubscribe_success
 question: |
-  You are now subscribed to reminders from "${ all_variables(special='titles').get('title', AL_ORGANIZATION_TITLE) }"
+  You are now subscribed to reminders from "${ all_variables(special='metadata').get('title', AL_ORGANIZATION_TITLE) }"
 subquestion: |
   You will continue to get any reminders of important dates by email or text message that you
   have not already been sent.
+back button: False  
 ---
 template: al_reminder_filing_template
 subject: |
-  Did you file your ${ all_variables(special='titles').get('title', AL_ORGANIZATION_TITLE) } documents yet?
+  Did you file your ${ all_variables(special='metadata').get('title', AL_ORGANIZATION_TITLE) } documents yet?
 content: |
   If you haven't filed yet, make a plan or do it today!
 ---
@@ -185,10 +189,10 @@ code: |
       if today() >= as_datetime(al_reminders[reminder]["date"]) and task_not_yet_performed(f"al_reminder_{reminder}"):
         if al_user_preferred_reminder_formats.get("email") and task_not_yet_performed(f"al_reminder_{reminder}_email"):
           if not send_email(to=al_user_reminder_email, template=value(al_reminders[reminder]["email template"]), task=f"al_reminder_{reminder}_email"):
-            log(f"{ all_variables(special='titles').get('title', AL_ORGANIZATION_TITLE) }: Couldn't send reminder to { showifdef('users[0]') } via {al_user_reminder_email}")
+            log(f"{ all_variables(special='metadata').get('title', AL_ORGANIZATION_TITLE) }: Couldn't send reminder to { showifdef('users[0]') } via {al_user_reminder_email}")
         if al_user_preferred_reminder_formats.get("sms") and task_not_yet_performed(f"al_reminder_{reminder}_sms"):
           if not send_sms(to=al_user_reminder_phone, template=value(al_reminders[reminder]["sms template"]), task=f"al_reminder_{reminder}_sms"):
-            log(f"{ all_variables(special='titles').get('title', AL_ORGANIZATION_TITLE) }: Couldn't send reminder to { showifdef('users[0]') } via {al_user_reminder_phone}")
+            log(f"{ all_variables(special='metadata').get('title', AL_ORGANIZATION_TITLE) }: Couldn't send reminder to { showifdef('users[0]') } via {al_user_reminder_phone}")
         mark_task_as_performed(f"al_reminder_{reminder}")
   al_reminders_cron_daily = True
 ---
@@ -210,10 +214,10 @@ code: |
   if al_user_wants_reminders and task_not_yet_performed(f"al_reminder_initial_reminder"):
     if al_user_preferred_reminder_formats.get("email") and task_not_yet_performed(f"al_reminder_initial_reminder_email"):
       if not send_email(to=al_user_reminder_email, template=value(al_reminders.initial_email_template), task=f"al_reminder_initial_reminder_email"):
-        log(f"{ all_variables(special='titles').get('title', AL_ORGANIZATION_TITLE) }: Couldn't send reminder to { showifdef('users[0]') } via {al_user_reminder_email}")
+        log(f"{ all_variables(special='metadata').get('title', AL_ORGANIZATION_TITLE) }: Couldn't send reminder to { showifdef('users[0]') } via {al_user_reminder_email}")
     if al_user_preferred_reminder_formats.get("sms") and task_not_yet_performed(f"al_reminder_initial_reminder_sms"):
       if not send_sms(to=al_user_reminder_phone, template=value(al_reminders.initial_sms_template), task=f"al_reminder_initial_reminder_sms"):
-        log(f"{ all_variables(special='titles').get('title', AL_ORGANIZATION_TITLE) }: Couldn't send reminder to { showifdef('users[0]') } via {al_user_reminder_phone}")
+        log(f"{ all_variables(special='metadata').get('title', AL_ORGANIZATION_TITLE) }: Couldn't send reminder to { showifdef('users[0]') } via {al_user_reminder_phone}")
     mark_task_as_performed(f"al_reminder_initial_reminder")
 
   al_reminders_sent_initial_test_message = True
--- a/docassemble/AssemblyLine/data/questions/al_reminders.yml
+++ b/docassemble/AssemblyLine/data/questions/al_reminders.yml
@@ -58,6 +58,22 @@ data:
     email template: al_reminder_filing_template
     sms template: al_reminder_filing_template
 ---
+code: |
+  al_reminders[i]["initial sms template"] = "al_reminder_initial_sms_template"
+  al_reminders[i]["initial email template"] = "al_reminder_initial_email_template"
+---
+template: al_reminder_initial_sms_template
+content: |
+  You asked to get a reminder of important dates when you used "${ all_variables(special='titles').get('title', AL_ORGANIZATION_TITLE) }".
+  You will get the reminders at this phone number.
+---
+template: al_reminder_initial_email_template
+subject: |
+  You are now signed up for reminders
+content: |
+  You asked to get a reminder of important dates when you used "${ all_variables(special='titles').get('title', AL_ORGANIZATION_TITLE) }".
+  You'll get the reminders using this email address.
+---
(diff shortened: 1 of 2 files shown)


docassemble/AssemblyLine/data/questions/al_reminders.yml
```diff
@@ -0,0 +1,89 @@
...
+content: |
+  If you haven't filed yet, make a plan or do it today!
+---
+event: cron_daily
+code: |
+  if al_user_wants_reminders:
+    import time
+    from random import random
+    # Avoid all of the interviews using up resources at the same time, vary the start time a bit
+    time.sleep(random())
```
BryceStevenWilley, line 73: Good idea, but is a < 1 second time variation enough? idk how long the email and sms tasks take, maybe more like max 10 or 20 seconds would be good. Guess it also depends on how many of these we end up having.

docassemble/AssemblyLine/data/questions/al_reminders.yml
```diff
@@ -0,0 +1,89 @@
...
+
+    for reminder in al_reminders:
+      if today() >= as_datetime(al_reminders[reminder]["date"]) and task_not_yet_performed(f"al_reminder_{reminder}"):
+        if task_not_yet_performed(f"al_reminder_{reminder}_email") and al_user_preferred_reminder_formats["email"]:
+          send_email(to=al_user_reminder_email, template=value(al_reminders[reminder]["email template"]), task=f"al_reminder_{reminder}_email")
+        if task_not_yet_performed(f"al_reminder_{reminder}_sms") and al_user_preferred_reminder_formats["sms"]:
+          send_sms(to=al_user_reminder_phone, template=value(al_reminders[reminder]["sms template"]))
+          mark_task_as_performed(f"al_reminder_{reminder}_sms")
```
BryceStevenWilley, line 81: `send_email` and `send_sms` both take a `task` param, it helps to be consistent there.

docassemble/AssemblyLine/data/questions/al_reminders.yml
```diff
@@ -0,0 +1,89 @@
...
+code: |
+  if al_user_wants_reminders:
+    import time
+    from random import random
+    # Avoid all of the interviews using up resources at the same time, vary the start time a bit
+    time.sleep(random())
+
+    for reminder in al_reminders:
+      if today() >= as_datetime(al_reminders[reminder]["date"]) and task_not_yet_performed(f"al_reminder_{reminder}"):
+        if task_not_yet_performed(f"al_reminder_{reminder}_email") and al_user_preferred_reminder_formats["email"]:
```
BryceStevenWilley, line 77: nit: I'd switch the `al_preferred_reminder_formats` check with the task check, makes slightly more sense reading, and doesn't require hitting a DB/another process each time.

BryceStevenWilley, line 77:
```suggestion
        if task_not_yet_performed(f"al_reminder_{reminder}_email") and al_user_preferred_reminder_formats.get("email"):
```

docassemble/AssemblyLine/data/questions/al_reminders.yml
BryceStevenWilley, line 81: I'm not sure how best to handle errors when sending the info. At the least outputting something to the log would be really helpful, otherwise we get a lot of silent errors.

BryceStevenWilley (commented):
Some minor nits, otherwise looks pretty good to me.

 A fairly big addition that might be useful would be to immediately send a message to the email / SMS like "Hi! We'll contact you at this number / email for your reminder in X days", just users can verify that they entered the right info, and don't get that unpleasant surprise the day after.

docassemble/AssemblyLine/data/questions/al_reminders.yml
nonprofittechy: I had ended up fine-tuning this performance on interviews.gbls.org. I had initially run into a big CPU spike when the cron jobs ran on a few hundred interviews. This seemed to be enough variation there, but I agree adding a few extra seconds is a good idea. I did up to 10 minutes originally but lowered it for manual tests.

docassemble/AssemblyLine/data/questions/al_reminders.yml
```diff
@@ -0,0 +1,89 @@
...
+        if task_not_yet_performed(f"al_reminder_{reminder}_sms") and al_user_preferred_reminder_formats["sms"]:
+          send_sms(to=al_user_reminder_phone, template=value(al_reminders[reminder]["sms template"]))
+          mark_task_as_performed(f"al_reminder_{reminder}_sms")
+        mark_task_as_performed(f"al_reminder_{reminder}")
+
+    if all(
+      task_performed(f"al_reminder_{reminder}")
+      for reminder in al_reminders
+    ):
+      allow_cron = False # Stop calling cron on this session once all tasks performed
```
plocket, line 88: Will this stop all cron tasks from being performed, or just the ones for `al_reminders`? I'd be concerned about the former since the developer might want other tasks to continue. Maybe an `all_al_reminders_complete` flag would be less disruptive?

plocket (commented):
I like this format in general as MVP. From what I can tell this will allow sending reminders to multiple people? I know it's something I would take advantage of. I'm not sure how it would be implemented, though, if it was for a dynamic number of people.

> A fairly big addition that might be useful would be to immediately send a message to the email / SMS...

An initial test message sounds like a great addition to me as well.

docassemble/AssemblyLine/data/questions/assembly_line.yml
```diff
@@ -7,6 +7,7 @@ include:
   - al_visual.yml
   - al_document.yml
   - ql_baseline.yml
+  - docassemble.ALToolbox:phone-number-validation.yml
```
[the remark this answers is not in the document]
BryceStevenWilley: Makes sense to move it just to keep things close together, but AssemblyLine already requires ALToolbox; `al_saved_sessions.yml` and `al_visual.yml` require it, and I think even if you don't use those features, DA will error if ALToolbox isn't installed.

docassemble/AssemblyLine/data/questions/al_reminders.yml
```diff
@@ -0,0 +1,89 @@
...
+    # Avoid all of the interviews using up resources at the same time, vary the start time a bit
+    time.sleep(random())
+
+    for reminder in al_reminders:
+      if today() >= as_datetime(al_reminders[reminder]["date"]) and task_not_yet_performed(f"al_reminder_{reminder}"):
+        if task_not_yet_performed(f"al_reminder_{reminder}_email") and al_user_preferred_reminder_formats.get("email"):
+          send_email(to=al_user_reminder_email, template=value(al_reminders[reminder]["email template"]), task=f"al_reminder_{reminder}_email")
+        if task_not_yet_performed(f"al_reminder_{reminder}_sms") and al_user_preferred_reminder_formats["sms"]:
+          send_sms(to=al_user_reminder_phone, template=value(al_reminders[reminder]["sms template"]))
+          mark_task_as_performed(f"al_reminder_{reminder}_sms")
```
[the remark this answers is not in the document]
BryceStevenWilley: `send_email` takes the `task` parameter that does the same thing. I think `send_sms` has the same param too.

docassemble/AssemblyLine/data/questions/assembly_line.yml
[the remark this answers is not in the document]
nonprofittechy: ALToolbox is already a dependency of the basic AL package (and also see #552)

docassemble/AssemblyLine/data/questions/al_reminders.yml
```diff
@@ -0,0 +1,89 @@
...
+    email template: al_reminder_filing_template
+    sms template: al_reminder_filing_template
+---
+template: al_reminder_filing_template
+subject: |
+  Did you file your ${ all_variables(special='titles').get('title', AL_ORGANIZATION_TITLE) } documents yet?
+content: |
+  If you haven't filed yet, make a plan or do it today!
+---
+event: cron_daily
```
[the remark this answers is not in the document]
nonprofittechy: This is a built-in event that we can't control the name of

docassemble/AssemblyLine/data/questions/al_reminders.yml
nonprofittechy: It would stop all cron from running. There are important reasons for this. See https://docassemble.org/docs/background.html#deleting: 

>Interviews containing scheduled tasks will run regularly, and the interview variables will be updated. This means that even if there is no activity from the original user, there will appear to be regular activity in the interview, which means that the [interview deletion](https://docassemble.org/docs/background.html#deleting) feature will never delete such interviews. Usually, it is a good thing that the [interview deletion](https://docassemble.org/docs/background.html#deleting) feature does not automatically delete interviews with scheduled tasks; you might have an interview that does something after a period of several months have passed.

Without the `allow_cron = False`, our promise to delete sessions after 180 days is a lie. It's important to tell the system when all cron tasks are done so the session can "age out" normally.

Unfortunately, the cron system is pretty inflexible. You can only have one `cron` event for each of `daily`, `hourly`, `weekly`, and `monthly`. But the benefit is someone can override our `cron_daily` event very easily. Only one block will run--the one latest in the YAML file using Docassmble's usual system for prioritizing a block.

nonprofittechy:
> Some minor nits, otherwise looks pretty good to me.
> 
> A fairly big addition that might be useful would be to immediately send a message to the email / SMS like "Hi! We'll contact you at this number / email for your reminder in X days", just users can verify that they entered the right info, and don't get that unpleasant surprise the day after.

It will have to be invoked by the developer separately, but I can add in some code for that.

docassemble/AssemblyLine/data/questions/al_reminders.yml
```diff
@@ -66,24 +66,32 @@ content: |
...
+            log(f"{ all_variables(special='titles').get('title', AL_ORGANIZATION_TITLE) }: Couldn't send reminder to { showifdef('users[0]') } via {al_user_reminder_phone}")
         mark_task_as_performed(f"al_reminder_{reminder}")
 
     if all(
       task_performed(f"al_reminder_{reminder}")
       for reminder in al_reminders
     ):
       allow_cron = False # Stop calling cron on this session once all tasks performed
+  else:
+    allow_cron = False
```
plocket, line 112: I feel like if this were at all a problem, it would have been spotted by someone else already, but how `allow_cron` is handled is niggling at me. We can wait to go into it till the next deep dive, though.

Details to discuss: I don't know enough about this part of the environment. It seems like putting it here would make order of execution integral to this process and more confusing to work out on another end. If it were, for example, a variable called `cancel_al_reminders`, then there could easily be a block that a developer using reminders can override. Putting `allow_cron` in there would remove the worries about order of execution?

docassemble/AssemblyLine/data/questions/al_reminders.yml
```diff
@@ -94,4 +110,16 @@ code: |
...
+  if al_user_wants_reminders and task_not_yet_performed(f"al_reminder_initial_{reminder}"):
+    if al_user_preferred_reminder_formats.get("email") and task_not_yet_performed(f"al_reminder_initial_{reminder}_email"):
+      if not send_email(to=al_user_reminder_email, template=value(al_reminders[reminder]["initial email template"]), task=f"al_reminder_initial_{reminder}_email"):
+        log(f"{ all_variables(special='titles').get('title', AL_ORGANIZATION_TITLE) }: Couldn't send reminder to { showifdef('users[0]') } via {al_user_reminder_email}")
+    if al_user_preferred_reminder_formats.get("sms") and task_not_yet_performed(f"al_reminder_initial_{reminder}_sms"):
+      if not send_sms(to=al_user_reminder_phone, template=value(al_reminders[reminder]["initial sms template"]), task=f"al_reminder_initial_{reminder}_sms"):
+        log(f"{ all_variables(special='titles').get('title', AL_ORGANIZATION_TITLE) }: Couldn't send reminder to { showifdef('users[0]') } via {al_user_reminder_phone}")
+    mark_task_as_performed(f"al_reminder_initial_{reminder}")
+
+  al_reminders_sent_initial_test_message
```
plocket, line 125: Is `al_reminders_sent_initial_test_message` supposed to be a variable used to trigger this block? If so, doesn't it need to be given a value here?

docassemble/AssemblyLine/data/questions/al_reminders.yml
nonprofittechy: Keep in mind there is only ever exactly one cron_daily event that can run, and that none of our partners are using cron at all right now.

Good to think of future flexibility but I think this might be too early a point to add anything like that.

BryceStevenWilley: Not super familiar with DA's cron system, but from everything in this PR, it seems like it's just not very compatible to being used by a library like ours, i.e. it's written as if everyone would be writing each interview entirely by hand, and not including as much cron code around. There might be some upstream fixes that could improve this compatibility, so I'm more towards getting something in over making it as flexible as possible too.

docassemble/AssemblyLine/data/questions/al_reminders.yml
```diff
@@ -87,12 +102,23 @@ code: |
...
+  response()
+---
+code: |
   if al_user_wants_reminders:
     import time
     from random import random
     # Avoid all of the interviews using up resources at the same time, vary the start time a bit
     time.sleep(random()*10)
-
+    log("Running cron")
```
plocket, line 121: Is this log here for development debugging, or is it meant to stay here? If it's meant to stay, could it be more specific to AL?

nonprofittechy: Just a temporary patch, it's not permanent. No need to review this commit-I had to save work while switching from work to home PC.

BryceStevenWilley:
A relevant piece of feedback for sms reminders: https://github.com/GBLS/docassemble-MAEvictionDefense/issues/239.

I think we thought that as long as we send just one SMS instead of multiple, we didn't need a technical way for folks to ask to stop the SMS (looks like MADE has multiple text reminders though), but it might be a good idea to let folks know this will be their only text reminder in the reminder itself, i.e. "This will be the last time we text for for case XYZ".

nonprofittechy:
> A relevant piece of feedback for sms reminders: [GBLS/docassemble-MAEvictionDefense#239](https://github.com/GBLS/docassemble-MAEvictionDefense/issues/239).
> 
> I think we thought that as long as we send just one SMS instead of multiple, we didn't need a technical way for folks to ask to stop the SMS (looks like MADE has multiple text reminders though), but it might be a good idea to let folks know this will be their only text reminder in the reminder itself, i.e. "This will be the last time we text for for case XYZ".

I'll have to check logs but this specific one makes me paranoid that multiple messages are going out due to a bug

PR #644, opened 2022-12-15
Better API for searching courts by county

### LINKED ISSUE #637: Create a more complete API for common uses of ALCourtLoader object
E.g.,:
1. Let people start with the county
2. 90% of the time (for smaller towns and cities),the county will let you get to a court without requiring you to make any more selection
3. If the county has more than one court, let people see the "full name" of the court. The full name might be composed of multiple potential columns in the spreadsheet. We need some API to let people say what columns they want to be part of the "full name" because people can add columns with whatever naming scheme they like.

nonprofittechy:
Most states have one court per county, except in the largest population counties. This adds to the ALCourtLoader API to make it very easy to figure out if the given county has multiple courts and adds a simple method of asking a follow-up question about the specific court the user wants to file in from among the list of courts in the given county.

I've also added a method that allows doing simple keyword searches against an arbitrary column to further filter the results. This is because in the very largest county, there are often specialty courts with limited jurisdiction--such as a family court, juvenile court, etc.

New methods:

* county_list - list all of the unique county names in the spreadsheet
* county_has_one_court(county) - true/false if the county has more than one court
* county_court(county) - returns the one court for the specified county without requiring you to lookup the court's index yourself
* matching_courts_in_county(county) - same as filter_courts, but just defaults to filtering by the county. This also lets you filter by an additional keyword if you want

Fix #637

Here's an example interview that uses the new method together with this list of courts in Minnesota:
[MinnesotaCourts.xlsx](https://github.com/SuffolkLITLab/docassemble-AssemblyLine/files/10239493/MinnesotaCourts.xlsx)

```yaml
---
include:
  - assembly_line.yml
---
modules:
  - .al_courts
---
objects:
  - the_courts: ALCourtLoader.using(filename="MinnesotaCourts.xlsx")
---
id: interview order
mandatory: True
code: |
  the_county
  my_court
  display_court
---
code: |
  if the_courts.county_has_one_court(the_county):
    my_court = the_courts.county_court("my_court", the_county)
  else:
    my_court = the_courts.as_court("my_court", court_index)
---
question: |
  What county do you live in?
fields:
  - Pick a county: the_county
    code: sorted(the_courts.county_list())
---
question: |
  What court do you want to file in?
subquestion: |
  There is more than one court in your county. Pick the court.
fields:
  - Pick a court: court_index
    code: the_courts.matching_courts_in_county(the_county)
    #code: the_courts.matching_courts_in_county(the_county, search_string="juvenile", search_columns="division")
---
event: display_court
question: |
  You selected "${ my_court }"
subquestion: |
  ${ my_court.address.block() }
```

### DIFF
--- a/docassemble/AssemblyLine/al_courts.py
+++ b/docassemble/AssemblyLine/al_courts.py
@@ -11,8 +11,9 @@ from docassemble.base.util import (
 )
 from docassemble.base.legal import Court
 import pandas as pd
+import numpy as np
 import os
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Optional, Union, Set, Iterable, Sized
 
 __all__ = [
     "ALCourt",
@@ -162,6 +163,10 @@ class ALCourtLoader(DAObject):
 
     Built around Pandas dataframe.
     """
+    def init(self, *pargs, **kwargs):
+        super().init(*pargs, **kwargs)
+        if not hasattr(self, "filename"):
+            self.filename = self.file_name # This spelling was a mistake but is everywhere
 
     # TODO: I think this design makes sense vs saving/storing ALL courts in the Docassemble session.
     # But we might want to at least cache data in Redis to reduce disk hits.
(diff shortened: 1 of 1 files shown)


docassemble/AssemblyLine/al_courts.py
```diff
@@ -203,16 +294,18 @@ def _load_courts(self):
         """
         Return list of courts
         """
-        if "/" in self.file_name:
-            to_load = path_and_mimetype(self.file_name)[0]
+        if not hasattr(self, "filename") and hasattr(self, "file_name"):
+            self.filename = self.file_name
```
BryceStevenWilley, line 298: 🙏🏻  I was going to ask for that change, but you already got to it, thanks!

docassemble/AssemblyLine/al_courts.py
```diff
@@ -11,8 +11,9 @@
 )
 from docassemble.base.legal import Court
 import pandas as pd
+import numpy as np
```
BryceStevenWilley, line 14: Doesn't seem to be used?

BryceStevenWilley (commented):
I haven't gotten to play around with this yet, but an open problem that I still see is what you can do with the data you get.

At some point in the interview, you'll want one specific court, but how you get from the filtered courts to that one court is awkward, since you need to pass around an index, etc.

Object labellers seem like they would work really once you're at the point of selecting a specific court: maybe matching courts in county or filter courts could return ALCourt objects?

Willing to defer to authors in the trenches here, but I feel like it could be a lot more powerful with the full court objects, not just a list of dicts for a dropdown.

PR #646, opened 2022-12-15
Add a parameter to always write out the full name of the state, instead of abbreviation

### LINKED ISSUE #296: Add a method on ALAddress to show the long form of the state/province/sublocality

nonprofittechy:
Fix #296 for real this time

Here is my test interview:

```yaml
---
include:
  - assembly_line.yml
---
mandatory: True
code: |
  users[0].address.address
  show_address
---
id: your address
sets:
  - users[0].address.address
  - users[0].address.city
  - users[0].address.zip
  - users[0].address.unit
  - users[0].address.state
  - users[0].address.country
question: |
  What is your address?
subquestion: |
  we are here
fields:
  - code: |
      users[0].address_fields(country_code=AL_DEFAULT_COUNTRY, default_state=AL_DEFAULT_STATE, show_country=False)
---
code: |
  AL_DEFAULT_COUNTRY = "NL"
---
event: show_address
question: |
  Here's your address both long and short
subquestion: |
  ${ users[0].address.block() }
  
  ${ users[0].address.block(long_state=True) }
  
  ${ users[0].address.line_two() }
  
  ${ users[0].address.line_two(long_state=True) }
  
  ${ users[0].address.on_one_line() }
  
  ${ users[0].address.on_one_line(long_state=True) }
```


docassemble/AssemblyLine/al_general.py
```diff
@@ -389,6 +440,22 @@ def normalized_address(self) -> Union[Address, "ALAddress"]:
             return self.norm_long
         return self
 
+    def state_name(self, country_code=None):
+        if country_code:
+            return state_name(self.state, country_code=country_code)
+        # Do a quick check for a valid ISO country code (alpha-2 or alpha-3)
+        if hasattr(self, "country") and self.country and len(self.country) < 4:
```
BryceStevenWilley, line 447: Does `alpha-3` actually work? Documentation just says alpha-2, but haven't checked source for that yet.

docassemble/AssemblyLine/al_general.py
```diff
@@ -87,7 +88,21 @@ def address_fields(
         `users[0].address.address`.
         Optionally, add a `show if` modifier to each field. The `show if` modifier
         will not be used if the `allow_no_address` modifier is used.
+
```
BryceStevenWilley, line 91: Should also mention what country codes are valid (maybe a link to https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#Officially_assigned_code_elements).

```suggestion
    `country_code` should be an ISO-3166-1 alpha-2 code (https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#Officially_assigned_code_elements)
    
```

BryceStevenWilley (approved):
LGTM

docassemble/AssemblyLine/al_general.py
nonprofittechy: Doesn't actually look like it will with the way that `state_name()` function is written, although pycountry does work with alpha_3 in general. This check just skips one try...except though.

Issue #320, opened 2021-10-20
Agree on a standard section and question order


nonprofittechy:
If we want to implement sections in the Weaver, per https://github.com/SuffolkLITLab/docassemble-ALWeaver/issues/138, it would help to have some consensus on standard question groupings, as well as the default order for some questions common to all interviews.

Here's my first draft:

1. Intro
2. Can I ... (get a restraining order? stop my eviction?) - e.g., a screening page
3. My name
3. My claim/defenses
4. My information
5. Information about the other side
6. Court selection
6. Review
6. Signature
7. Download and next steps

It might be a really nice project to have someone (student?) audit some popular forms on LHI to see what sections are used by high quality guided interviews.

nonprofittechy:
See https://github.com/SuffolkLITLab/form-explorer/issues/20 for a way to achieve in the Weaver

nonprofittechy:
This sparked a good discussion, but there is already some nascent code in the Weaver, which is where the rest of the discussion will end up happening. We also have a [document](https://suffolklitlab.org/docassemble-AssemblyLine-documentation/docs/style_guide/question_structure) on this.

PR #647, opened 2022-12-15
Add a link to "original form" on the default about page, if it exists and looks like a valid link

### LINKED ISSUE #260: Feedback from Appeal Form: Overview map wanted
From this bit of feedback: https://github.com/SuffolkLITLab/docassemble-AppealsCivilDocketingStatement/issues/21#issue-956957478

> it would be helpful to offer the full and complete form in blank so user either can download and print that to complete manually or with other applications, OR, so user can see where the form Q&A is heading and know what is being requested so they give best answers

1. Are we linking to the original form in the About page yet / should we? 
2. Some sort of overview map might make sense? Progress bar doesn't really fit for what this person wants. I feel we could make this feature with the DA navigation stuff, but it's not super clear if that would be enough. 

Far future, it would be cool to do a Docassemble look ahead: where you can see all of the possible next screens, in some sort of DAG. But that's much more than we need here, just thought it was a cool idea.

nonprofittechy:
This covers a suggestion in #260 to give users a little more context in their progress on the form. It doesn't add sections to the interview, but we already have open issues in the Weaver for that (See https://github.com/SuffolkLITLab/docassemble-ALWeaver/issues/45), which is really the only place we can help authors with this task.

Also see #320

Fix #260

### DIFF
--- a/docassemble/AssemblyLine/data/questions/al_visual.yml
+++ b/docassemble/AssemblyLine/data/questions/al_visual.yml
@@ -333,6 +333,22 @@ template: about_this_interview_version_info
 content: |
   "${all_variables(special='metadata').get('title','').rstrip()}" version 
   `${ package_version_number }`; AssemblyLine version `${ al_version }`.
+
+  % if interview_metadata.get("main_interview_key"):
+    <%
+      MAIN_METADATA = interview_metadata[interview_metadata["main_interview_key"]]
+    %>
+  % elif len(interview_metadata) > 1:    
+    <% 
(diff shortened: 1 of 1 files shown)


BryceStevenWilley:
Will need to get added to [MassAccess](https://github.com/SuffolkLITLab/docassemble-MassAccess), which overrides this for MA interviews (tried testing for a bit, wondering why it wasn't showing up).

BryceStevenWilley (approved):
LGTM!

PR #652, opened 2023-01-03
Cleanup appearance of error action page

nonprofittechy:
- Use better formatting to more closely match existing DA 501.html template
- Make it possible to completely hide the technical information when verbose errors are disabled with the existing Docassemble error configuration (https://docassemble.org/docs/config.html#verbose%20error%20messages)

Test interview:

```yaml
---
include:
  - assembly_line.yml
---
mandatory: True
question: |
  
  ${ users[dict()] }
```

### DIFF
--- a/docassemble/AssemblyLine/data/questions/al_visual.yml
+++ b/docassemble/AssemblyLine/data/questions/al_visual.yml
@@ -495,12 +495,14 @@ content: |
   <h2 class="h4">${ word("Error") }</h2>
   ${ action_argument('error_message') }
 
-  % if get_config("debug") and action_argument("error_trace"):
+  % if get_config("debug") and action_argument("error_history"):
   <h2 class="h4">${ word("History") }</h2>
   <pre>
   ${ action_argument("error_history") }
   </pre>
+  % endif
 
+  % if get_config("debug") and action_argument("error_trace"):
   <h2 class="h4">${ word("Log") }</h2>
   <pre>
   ${ action_argument("error_trace")}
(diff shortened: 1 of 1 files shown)


docassemble/AssemblyLine/data/questions/al_visual.yml
```diff
@@ -448,7 +448,9 @@ subquestion: |
   * ${ al_opt }
   % endfor
 
+  % if not get_config("verbose error messages") == False:
```
BryceStevenWilley, line 451: Should this be under `assembly line`? Not sure what we've been doing there, I do know there's an open issue for it.

BryceStevenWilley (commented):
LGTM. Did we have a specific ask for not showing detailed error info?  I'm okay with being added I guess, but IMO, it's not productive: the detailed info isn't likely to be distracting to users or leak any PI or interview info (everything is open source anyway) and it only makes things harder for developers (us) when trying to reproduce or fix bugs from users.

docassemble/AssemblyLine/data/questions/al_visual.yml
nonprofittechy: No, this is an existing Docassemble configuration option that I figured we should respect. It defaults to `True` but it's not defined in the normal config which is why we're doing an explicit check for `False`. https://docassemble.org/docs/config.html#verbose%20error%20messages

BryceStevenWilley: Gotcha, that's fair.

BryceStevenWilley:
After realizing that this is a bit more aligned with a DA config, I get it a bit more. I still think that particular config isn't well thought out (with it on, error screens just show the word "error" and nothing else at all), and I wouldn't encourage anyone to enable it, but it we want to be aligned with it we can be.

BryceStevenWilley:
I added a commit that splits up the trace and the history, which can appear separately (i.e., a history with no trace will show up when there's an infinite loop).

PR #667, opened 2023-03-22
Actually print the email_to_use in the email code

BryceStevenWilley:
Still needs testing, which can be a bit intensive to setup.


nonprofittechy:
@BryceStevenWilley I think we can close this, right? Now that we're live with e-filing?

BryceStevenWilley:
I did close some of the other ones last week, but I realized on Friday that we do actually still use the email code; the Appeals court specifically asked us to email them when we make an e-filing, since it's still useful for emergency cases. Tested that it worked on Friday as well.

The emailing code does become less important now that e-filing is out though, so once I launch everything today, I can just merge this, without super extensive testing.

PR #668, opened 2023-03-27
Add fields to use in name change interviews

nonprofittechy:


docassemble/AssemblyLine/data/questions/ql_baseline.yml
```diff
@@ -2008,3 +2008,206 @@ generic object: ALAddress
...
+id: consent of parent 1
+question: |
+  Do you consent to ${ children[0].preferred_name }'s name change?
+fields:
+  - I consent: users[0].consented_to_name_change
+    datatype: yesnoradio
+---
+id: consent of parent 1
+question: |
+  Does ${ users[1] } consent to ${ children[0].preferred_name }'s name change?
```
BryceStevenWilley, line 2184: I don't think the consent questions belong in AL like this; they're very specific to a single for or set of forms, and I can't imagine anyone using them without modifying something about them.

docassemble/AssemblyLine/data/questions/ql_baseline.yml
```diff
@@ -2008,3 +2008,206 @@ generic object: ALAddress
...
 content: |
   Example: "I normally sleep at 5th and Main."
+---
+################## Name change questions ######################
+---
+id: What is your new name?
+question: |
+  What name do you want to be called?
+fields:
+  - First name: users[0].preferred_name.first
```
BryceStevenWilley, line 2018: The only issue I have here is the repeated use of the thing that we already have; the `name_fields()`. I'm having trouble thinking of better ways to organize it, but I think it's worth considering if we should refactor and move `name_fields()` to the name object instead of the person object. That's pretty hard though, and I am afraid it would break a lot of things.

plocket: Could just keep the current behavior, deprecate it, and add `name_fields()` to the name object if we want a longer transition time or to grandfather the old behavior in while establishing a new convention.

PR #684, opened 2023-04-16
Document bundle compatibility

### LINKED ISSUE #683: ALDocumentBundles are missing some methods of ALDocument class
E.g., missing the is_docx() method which results in this bug w/ the new Affidavit of Indigency:

https://github.com/nonprofittechy/docassemble-HousingCodeChecklist/issues/357

nonprofittechy:
Fix #683 

We released the new Affidavit of Indigency as an ALDocumentBundle, not a plain ALDocument anymore. I forgot to make sure it worked in all scenarios; it doesn't work when you show a download screen with DOCX downloads because of the missing is_docx() method.

I went through and tried to implement a basic version of all of the ALDocument methods on ALDocumentBundle so we can try to avoid other issues like this in less commonly tested branches.

### DIFF
--- a/docassemble/AssemblyLine/al_document.py
+++ b/docassemble/AssemblyLine/al_document.py
@@ -31,6 +31,7 @@ from docassemble.base.pdfa import pdf_to_pdfa
 from textwrap import wrap
 from math import floor
 import subprocess
+from collections import ChainMap
 
 __all__ = [
     "ALAddendumField",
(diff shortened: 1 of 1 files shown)


nonprofittechy:
> I like `_is_docx`, and `as_docx`, those are pretty neat. But I don't know if I understand why we'd add `safe_value()`, `overflow_value()`, etc.
> 
>     * what's the intended use? From what I remember, things like `overflow()`, and `overflow_value()` are used in the addendum docxs, but you won't be able to access the ALDocumentBundle from the addendum, only the ALDocument itself?

The purpose is to let someone treat something like the Affidavit of Indigency as a single document and avoid syntax errors if they mistakenly reference the wrong item. Agreed those are not as likely to get used.

>     * how are these functions defined if the same variable is used in two different documents, but has a different safe value for each?

There is one shared interview variable space. You can't have variables with separate values per document. But you are right that if the documents repeat the same variables with different amounts of space on the PDF, the safe value would be different. The fields that get sent to an addendum are longer text fields and it would be unusual for them to get repeated.

If there is ambiguity, the author would need to drop down a level to the actual ALDocument contained in the bundle.

> I do like trying to make a concrete shared interface between ALDocument and ALDocumentBundle, but I think it should be a subset of the ALDocument interface, not all of it.

I had never considered the `_is_docx()` issue would come up before either. I'm thinking some future proofing is smart here rather than rushing to fix the next unanticipated error. But maybe this just isn't likely enough to add code for...

BryceStevenWilley:
> The purpose is to let someone treat something like the Affidavit of Indigency as a single document and avoid syntax errors if they mistakenly reference the wrong item. Agreed those are not as likely to get used.

That's fair, but again, I think the best way to handle this is to make a shared interface between ALDocument and all of it's friends, and then write interviews such that they interact with that class, and not ALDocument directly. A lot of the complex ways of dealing with ALDocuments are in the `al_document.py` code, which we can type check. I have a quick in-progress branch for that.

> I'm thinking some future proofing is smart here rather than rushing to fix the next unanticipated error.

From my view, this PR is rushing to fix the next anticipated error. Future proofing would be structuring the classes to handle any changes between the document types (`ALDocument` -> `ALStaticDocument`, `ALDocumentBundle` -> `ALExhibitDocument`, etc.) without breaking the interview. There are other issues with the function signatures too, including the fact that ALDocumentBundle can return None for `as_pdf`, and one path in download_list_html assumes that we can the key we pass to ALDocumentBundle is the same index as sub-document bundles (it won't be), both of which break assumptions we have elsewhere.

I'm happy to merge the docx changes now, and put the overflow fields into a new PR, where we can also make sure the rest of the classes are structured well, and see if we need to expose those APIs at the "Abstract Document" level.

nonprofittechy:
> I'm happy to merge the docx changes now, and put the overflow fields into a new PR, where we can also make sure the rest of the classes are structured well, and see if we need to expose those APIs at the "Abstract Document" level.

That seems like a good approach. Thanks for thinking this through.

BryceStevenWilley:
Sorry, we were working on this at the same time, you merged right before I was about to push the same changes. For what it's worth, I put the overflow methods and the mypy typing I had been working on in https://github.com/SuffolkLITLab/docassemble-AssemblyLine/tree/overflow_additions, just so it's not lost

PR #687, opened 2023-05-08
Experimenting with not running Kiln on non-functional commits

BryceStevenWilley:
Tests shouldn't need to run when change the README or the CHANGELOG.

### DIFF
--- a/.github/workflows/run_form_tests.yml
+++ b/.github/workflows/run_form_tests.yml
@@ -10,7 +10,7 @@ on:
       - 'requirements.txt'
       - '**.py'
       - '**.yml'
-      - 'docassemble/AssemblyLine/**'
+      - 'docassemble/**'
   workflow_dispatch:
     inputs:
       tags:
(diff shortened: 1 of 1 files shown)


nonprofittechy:
Does this stop kiln from re-running when you modify a .feature file? is that desired?

nonprofittechy (changes requested):
Not sure this is a desirable or correct change

BryceStevenWilley:
> Does this stop kiln from re-running when you modify a .feature file?

No, this line should keep it running on anything in the actual package: `'docassemble/AssemblyLine/**'`. The `*.py` should keep it running on `setpy.py` or `__init__.py` files, that could break things (but really shouldn't).

.github/workflows/run_form_tests.yml
```diff
@@ -6,6 +6,11 @@ on:
       - "**"
     tags-ignore:
       - "**"
+    paths:
+      - 'requirements.txt'
+      - '**.py'
+      - '**.yml'
+      - 'docassemble/AssemblyLine/**'
```
nonprofittechy, line 13:
```suggestion
      - 'docassemble/**'
```

PR #700, opened 2023-06-07
add questions for a user's pronouns

### LINKED ISSUE #699: Add a base library question for pronouns
Something that the Appeals court has specifically expressed interest in.

Would need to decide on:
* standard variable name and location (likely attached to ALIndividuals, i.e. `users[0].pronouns`)
* draft language of the question (pretty sure we have similar questions somewhere, idk where though, searching "pronouns" in our org doesn't come up with any YAML)
* attach to existing name question (where it's always shown / we always have) or on a new question (too separate?)
* actually put it in the YAML

nonprofittechy:
Fix #699 

This adds a `pronoun_fields()` method that accepts these parameters:

* show_help: adds a help bubble that explains what pronouns are
* show_if: make the whole question conditional
* required: defaults to False
* shuffle: whether the order of pronouns is randomized

The response includes 5 sets of pronouns:

1. she/her/hers
2. he/him/his
3. they/them/theirs
4. ze/hir/hirs
5. ze/zir/zirs

As well as the options:

1. unknown
2. prefer not to say
3. something else

The accompanying `get_pronouns()` method returns the selected pronouns as a set as well as any self-described pronouns the user opts to use.

### DIFF
--- a/docassemble/AssemblyLine/al_general.py
+++ b/docassemble/AssemblyLine/al_general.py
@@ -26,6 +26,7 @@ from docassemble.base.util import (
     validation_error,
     word,
 )
+import random
 import re
 import pycountry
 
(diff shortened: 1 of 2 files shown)


nonprofittechy:
Going to merge--this is a new optional feature, I've tested all of the small number of combinations of uses, and it's something the Appeals Court is looking for in the next week or two (which seems definitely feasible to do).

PR #715, opened 2023-07-25
Load languages by default

### LINKED ISSUE #218: Identify a variable/flag to tell AL that the interview is available in multiple languages, and the list of languages
This probably goes in the interview_metadata object?

Benefits of metadata key: you can configure in each individual interview. Drawback: more complex to decide which interview's translation status controls. We may not want to show any translated text unless the umbrella interview is fully translated (users can find it confusing to read partially translated content).

Maybe we don't use the `interview_metadata` key, but just the regular `metadata` block?

### LINKED ISSUE #590: Integrate downstream translations
We have a few projects that have translated the ql_baseline.yml strings. We should think about a way to add those translations into this upstream package so they don't have to get re-translated.

nonprofittechy:
Fix #590  fix #218 

Add an interface to use a simple translation system without breaking existing interviews that use translation, and partial translations of Assembly Line phrases for Spanish, Portuguese, and Haitian Creole.

It provides three methods for the user to define their language:

1. A dropdown menu in the navigation bar
2. An inline list, which might be appropriate to include on the first page of the interview so it's more prominent
3. A URL parameter "lang" which is read the first time the interview session is started

The idea is for the user to click on the menu item or list item that matches their language, so this also incorporates a list of languages that have been translated into the native tongue as a YAML file. The interview author can point to a YAML file of their choice instead. I used ChatGPT to generate this list and stopped at 47 languages that might be seen in US immigrant populations. There's no definitive version of this list in something like pycountry.

Adds a few new AL prefixed magic variables:

* `enable_al_language`: defaults to True, turning it off can partially ensure the language system in AssemblyLine doesn't interfere with existing language systems. It should be relatively easy for authors to migrate to this new system though.
* `al_user_default_language`: can be controlled by interview author, this determines the language when the user makes no selection of their own. Defaults to "en".
* `al_interview_languages`: a list of language codes, presumably ISO-639-1 (Alpha-2), like ["en","es"] etc. They should exist in the languages.yml file if you would like to display in the native language. The author can provide their own languages.yml file.
* `al_user_language`: normally set via a dropdown menu or passed as a URL argument, this stores the user's current selected language

This has a chance to break existing multilingual interviews in some special circumstances (essentially if the author relies on a different default language and didn't set the default language in an initial block--that can be prevented by setting `enable_al_language` to `False`). I think a minor version number increment should be sufficient though--it would require some unique combination of factors. I don't think this will break any existing multilingual interviews I have worked on, and I'm not aware of other Assembly Line users who have multilingual interviews.


docassemble/AssemblyLine/data/sources/languages.yml
```diff
@@ -0,0 +1,141 @@
...
+  name: Telugu
+  native_name: తెలుగు
+uz: 
+  name: Uzbek
+  native_name: Oʻzbek
+pa: 
+  name: Punjabi
+  native_name: ਪੰਜਾਬੀ
+fa: 
+  name: Persian (Farsi)
```
BryceStevenWilley, line 119:
```suggestion
  name: Farsi (Persian)
```

Fairly minor, and up to debate, but Farsi is more specific than Persian, and might be more familiar to our audience of the Iranian diaspora

docassemble/AssemblyLine/data/sources/languages.yml
```diff
@@ -0,0 +1,141 @@
...
+  native_name: Ελληνικά
+ht: 
+  name: Haitian Creole
+  native_name: Kreyòl ayisyen
+zh-Hans: 
+  name: Mandarin (Simplified)
+  native_name: 普通话
+zh-Hant: 
+  name: Cantonese (Traditional)
+  native_name: 粵語
```
BryceStevenWilley, line 60:
```suggestion
zh-Hans: 
  name: Mandarin (Simplified)
  native_name: 简体中文
zh-Hant: 
  name: Mandarin (Traditional)
  native_name: 繁體中文
zh-HK:
  name: Cantonese (Simplified)
  native_name:  粤语 (简体)
```

ChatGPT def got this wrong; I replaced it with the characters used in Android's language selection.  But there are a few things we should figure out:

* Cantonese has a few different language codes; `yue` in ISO 639-3 seems to be the best one, but [zh-HK](https://sites.psu.edu/symbolcodes/languages/asia/chinese/) does seem to be recommended for some. Neither distinguish between traditional and simplified script, even though that is distinguished in the Android selection that I checked.
* If this list is primarily for choosing a language, I suggest that we remove `Chinese` from the top, and use these more specific ones, as they're a bit confusing, and mostly mean the same thing.  If we do, we might want to use `Chinese` and not `Mandarin`

BryceStevenWilley (commented):
Temporary review of just the language codes, I'll try to look at the rest tomorrow.

docassemble/AssemblyLine/data/sources/languages.yml
nonprofittechy: This is fine, but note we never show the English name of the language

docassemble/AssemblyLine/data/sources/languages.yml
nonprofittechy: We should feel free to have as many unique variations as we want. The author specifies the list of language codes that they want to use and this is just a lookup table for the native language version of the name. So I don't suggest we remove "Chinese," especially as the written form of Chinese is distinct and can be either traditional or simplified. Most Cantonese speakers use the traditional script and most Mandarin speakers us the simplified, but it's not totally 1 to 1.

In my existing interviews I use zh-t for traditional and I let users do Google Translate for switching between the two. Unlike character based languages, Chinese can be 100% machine translated, at least from info I received from the Asian Outreach Unit at GBLS. So it's less likely that an author will translate one interview into both systems.

The goal with the YAML is to have a good list of potential ISO/custom codes that authors might use.

nonprofittechy: Also found this reference: https://stackoverflow.com/questions/18902072/what-standard-do-language-codes-of-the-form-zh-hans-belong-to

BryceStevenWilley: > So I don't suggest we remove "Chinese," especially as the written form of Chinese is distinct and can be either traditional or simplified

This statement seems contradictory; because there are two different forms, I don't think we should the use a label that's more vague, rather than one of the two, more specific, labels. Like you said, we want "to have as many unique variations as we want", but `Chinese` by itself isn't really unique, it's an umbrella for a lot of variation.

That does make sense that someone won't translate one interview into both, so that's fair. But there's benefit in the label matching which translation was used.

FWIW, this is from the [RFC mentioned in your last link](https://www.rfc-editor.org/rfc/pdfrfc/rfc5646.txt.pdf)

> The language subtag ’zh’ can still be used without an extended language subtag to label a resource as some unspecified variety of Chinese, while the primary language subtag (’gan’, ’yue’, ’cmn’) is preferred to using the extended language form ("zh-gan", "zh-yue", "zh-cmn")

So, should probably use `zh-cmn-Hant`, `zh-cmn-Hans`, and `zh-yue`, maybe with extensions as well.

BryceStevenWilley (approved):
LGTM, I still think we should remove "Chinese" in favor of the more specific ones; in the python way, there should be one clear option for authors to choose for their translation.

docassemble/AssemblyLine/data/sources/languages.yml
nonprofittechy: It doesn't make a lot of sense AFAIK to talk about the written language (which is what Docassemble offers) as either Mandarin or Cantonese. So I think the `cmn` and `yue` would add confusion. That said I don't speak any Chinese language so I could be totally off base.

nonprofittechy: Thinking this over, I believe the description should be Chinese (simplified) and Chinese (traditional) and the tags should be zh-Hans and zh-Hant, without mentioning Cantonese or Mandarin.

nonprofittechy:
> LGTM, I still think we should remove "Chinese" in favor of the more specific ones; in the python way, there should be one clear option for authors to choose for their translation.

Maybe worth clarifying what I meant about having multiple options there. I was envisioning that sometimes folks would be stuck using an externally provided language code. In that situation if they want to use our drop-down language selector, I'd like them to be able to use the code in their externally provided file without altering it.

It's probably not a big deal for someone to change it in most circumstances. That said, I can see someone using a words.yml someone else generated that uses `zh` for Chinese, and since they just have one Chinese language they're offering, they generate their own translation files to match and choose `zh` as well.

It's also possible we'll end up creating a defacto standard.

docassemble/AssemblyLine/data/sources/languages.yml
BryceStevenWilley: That's probably fine. I had been basing a lot of my suggestions on the language options present in Android, but after looking at what Windows and Ubuntu have for language selections, I think you're right. I guess Android also has some spoken portions that would necessitate a difference between Cantonese and Mandarin. Sorry!

BryceStevenWilley:
> Maybe worth clarifying what I meant about having multiple options there. I was envisioning that sometimes folks would be stuck using an externally provided language code. In that situation if they want to use our drop-down language selector, I'd like them to be able to use the code in their externally provided file without altering it.

Sorry, I thought it would be something that the authors would have full control over from when I was playing around with this PR myself.

I was always pushing for distinguishing between the two since it's just a good practice that IMO helps set expectations with users. I'd be happy if we listed `zh` after `zh-Hans` and `zh-Hant` in the YAML, maybe adding a comment to prefer the others.

nonprofittechy:
> > Maybe worth clarifying what I meant about having multiple options there. I was envisioning that sometimes folks would be stuck using an externally provided language code. In that situation if they want to use our drop-down language selector, I'd like them to be able to use the code in their externally provided file without altering it.
> 
> Sorry, I thought it would be something that the authors would have full control over from when I was playing around with this PR myself.
> 
> I was always pushing for distinguishing between the two since it's just a good practice that IMO helps set expectations with users. I'd be happy if we listed `zh` after `zh-Hans` and `zh-Hant` in the YAML, maybe adding a comment to prefer the others.

I like this solution. It allows us to deal with existing translation files in the wild, while also recognizing that people might read our YAML file to make their own decisions about which languages to offer.

PR #718, opened 2023-07-26
Use "it" for a business, correctness for pronoun methods and handle preferred pronouns

### LINKED ISSUE #672: Allow user to modify how pronouns of person_type == 'business' are handled.
I read in the DA documentation that:
If the object is a generic DAObject, or a subclass of DAObject that does not have any special pronoun behavior, then the .pronoun(), .pronoun_subjective(), and .pronoun_objective() methods all return 'it'. The .pronoun_possessive() method returns 'its' followed by the argument.

I am trying to figure out how to modify my code so that the python possessive functions [pronoun(), pronoun_possessive(), pronoun_objective(), and pronoun_subjective()] when dealing with business entities as the object, function in the manner I desire. The first two situations below function as desired, it is the third scenario that is troubling me.

Current jinja2 code:
{{ (other_parties.pronoun_possessive() }}, {{ other_parties.pronoun_objective() }}, {{ other_parties.pronoun_subjective() }}
1. Desired output if other_parties contains more than one object:
their, them, they
Current (and correct) output:
their, them, they
2. Desired output if other_parties contains just one object with a male gender, and object is person_type != "business":
his, him, he
Current (and correct) output:
his, him, he
3. Desired output if other_parties contains just one object, and that object is person_type == "business":
its, it, it
Current (incorrect - based upon my personal preference) output:
their, them, they

------

Quinten Steenhuis
  [30 minutes ago](https://docassemble.slack.com/archives/C7791ATUJ/p1680267087247659?thread_ts=1680266963.728849&cid=C7791ATUJ)
This is part of the ALIndividual class from Suffolk and sounds like a good feature request. Want to open an issue here: https://github.com/SuffolkLITLab/docassemble-AssemblyLine/issues

nonprofittechy:
Fix #672 

I also realized Jonathan's original code had some mistakes, including a default "he/him/his" when gender wasn't one of "male", "female" or "other."

Finally, this code handles the user's preferred pronouns in priority to the pronouns that come from the default of their gender if the author has a question that asks for pronouns (and we can unambiguously parse those pronouns)

### DIFF
--- a/docassemble/AssemblyLine/al_general.py
+++ b/docassemble/AssemblyLine/al_general.py
@@ -1039,7 +1039,7 @@ class ALIndividual(Individual):
         )
 
     def pronoun(self, **kwargs):
-        """Returns a pronoun as appropriate, based on attributes.
+        """Returns an objective pronoun as appropriate, based on attributes.
 
         The pronoun could be "you," "her," "him," "it," or "them". It depends
         on the `gender` and `person_type` attributes and whether the individual
--- a/docassemble/AssemblyLine/al_general.py
+++ b/docassemble/AssemblyLine/al_general.py
@@ -38,29 +38,31 @@ import re
 import pycountry
 
 __all__ = [
+    "Abuser",
+    "Abuser",
+    "AddressList",
     "ALAddress",
     "ALAddressList",
-    "ALPeopleList",
     "ALIndividual",
-    "section_links",
-    "is_phone_or_email",
-    "section_links",
-    "Landlord",
-    "Tenant",
-    "HousingAuthority",
+    "ALPeopleList",
     "Applicant",
-    "Abuser",
-    "Survivor",
-    "PeopleList",
-    "VCIndividual",
-    "AddressList",
     "Applicant",
-    "Abuser",
-    "Survivor",
     "github_modified_date",
-    "will_send_to_real_court",
-    "safe_subdivision_type",
+    "has_parsable_pronouns",
+    "HousingAuthority",
+    "is_phone_or_email",
+    "Landlord",
     "language_name",
+    "parse_custom_pronouns",
+    "PeopleList",
+    "safe_subdivision_type",
+    "section_links",
+    "section_links",
+    "Survivor",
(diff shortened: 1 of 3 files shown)


docassemble/AssemblyLine/al_general.py
```diff
@@ -1031,6 +1038,160 @@ def address_block(
             )
         )
 
+    def pronoun(self, **kwargs):
+        """Returns a pronoun as appropriate, based on attributes.
```
BryceStevenWilley, line 1042:
```suggestion
        """Returns an objective pronoun as appropriate, based on attributes.
```

I'm okay with objective being the default, but it should be clearer that it is.

docassemble/AssemblyLine/al_general.py
```diff
@@ -1031,6 +1038,160 @@ def address_block(
...
+            str: The appropriate pronoun.
+        """
+        if self == this_thread.global_vars.user:
+            output = word("you", **kwargs)
+        elif (
+            hasattr(self, "pronouns")
+            and isinstance(self.pronouns, DADict)
+            and len(self.pronouns.true_values()) == 1
+            and self.pronouns.true_values()[0]
+            in ["she/her/hers", "he/him/his", "they/them/theirs", "ze/zir/zirs"]
```
BryceStevenWilley, line 1064: This is a super great structure IMO, but I think a really good improvement would be to accept any pronouns that have two slashes in them. `he/him/his` is pretty standard (I have seen slashes for people who are okay with multiple sets, like (she/they), but it's usually only two), and it doesn't feel great to have the option for users to enter their own pronouns, but then to ignore them.

BryceStevenWilley (approved):
LGTM! It'd be really nice to support people's self-filled in pronouns if they use the same formatting as we do, ti should fit in well there.

docassemble/AssemblyLine/al_general.py
nonprofittechy: You're right, I was just getting lazy. I'll add that before merging.

PR #726, opened 2023-08-08
More flexible API for menu items

### LINKED ISSUE #723: Is the method of adding menu items compatible with interviews that want to add their own extra items?
I think the `reconsider: True` breaks ability to add custom items. Need to revisit this

nonprofittechy:
fix #723

Adds a new user-facing magic keyword: `al_menu_items_custom_items` which should be a list of menu items. Custom items are shown before default menu items.

Each menu items follows the Docassemble menu item format, which is a dictionary with the keys `url` and `label`. It can be convenient, and readable, to use a `data from code` block to define the menu items, but you can use any method to create the right Python datastructure.

Optionally, you can add a key `hidden` which contains a True/False value. If the value is `True` the menu item won't be shown. This can be used to dynamically show or hide the menu item on a given page. If the value is dynamic, you should add a `reconsider` modifier on the block that defines `al_menu_items_custom_items`.

E.g.,

```
reconsider: True
variable name: al_menu_items_custom_items
data from code:
  - url: |
      "http://example.com"
    label: |
       "Example link"
    hidden: user_turned_off_example_link
```

Note that if the visible state isn't dynamic, the author can use a `data` block instead of a `data from code` block. This will make the label translatable. Alternatively, the author can add a template for the label to make it available for translation.

### DIFF
--- a/docassemble/AssemblyLine/data/questions/al_visual.yml
+++ b/docassemble/AssemblyLine/data/questions/al_visual.yml
@@ -69,54 +69,68 @@ features:
 ---
 initial: True
 code: |
-  # Set the dropdown/hamburger menu items
-  # by default we just add a "Start over" link
-  if _internal.get('steps') > 1:
-    menu_items = al_menu_items
-  else:
-    # Skip the "start over", download, and exit and erase menu item
-    if al_enable_incomplete_downloads:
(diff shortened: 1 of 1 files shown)


BryceStevenWilley (approved):
LGTM. Getting this in the docs would ne great too

Issue #461, opened 2022-05-02
Adjust vertical alignment when using a `decoration`


nonprofittechy:
![image](https://user-images.githubusercontent.com/7645641/166307315-b26d6665-c162-4e39-80de-27d6840f5e1c.png)

When you add a decoration to a `question`, it makes the text float up pretty far above the horizontal rule that separates `question` from `subquestion`. It would be nice if there were a CSS fix that would make the relative position of icon, horizontal rule, and `question` text more visually organized.

nonprofittechy:
General approaches:

1. Add a `span` around the text so it's its own element
2. We can then add vertical-alignment `middle` 
3. also reduce the size from 2em to 1.5 em

OR just fix it the "right" way with flexbox but that would likely break other people's heavily styled interviews

nonprofittechy:
I think we should really advocate strongly to fix this upstream. One of the handful of broken HTML issues in Docassemble that is kept around for questionable backwards compatibility IMHO.

PR #737, opened 2023-08-17
remove icon from intro screen that was never displayed

### LINKED ISSUE #736: Missing `sign-form-lineal.svg`
Lots of logs on production that say: `File reference docassemble.AssemblyLine:sign-form-lineal.svg DID NOT EXIST.` Idk what could be going on

https://github.com/SuffolkLITLab/docassemble-AssemblyLine/blob/ea9e2a08105265db6a93a25c40889ebc3da29d58/docassemble/AssemblyLine/data/questions/al_visual.yml#L195

nonprofittechy:
Fix #736

Looks like this was meant to be a decoration for the main intro page. I found the icon w/ the attribution, but it's now behind a paywall so we can't easily just add it back in anyway.

It's not a bad idea to have some visual here, but given #461 and the fact that this decoration appears to have *never* been visible, going to just remove it for now as the safest choice.

We can decide to add more graphics later, probably shouldn't do it in the default theme without notice though.

### DIFF
--- a/docassemble/AssemblyLine/data/questions/al_visual.yml
+++ b/docassemble/AssemblyLine/data/questions/al_visual.yml
@@ -185,15 +185,6 @@ default screen parts:
     [:info-circle: About](${ url_action('about_this_interview') })
     [:comment-dots: Feedback](${ feedback_link(user_info(), i=feedback_form, github_repo=github_repo_name, github_user=github_user, package_version=package_version_number) } ){:target="_blank"}
 ---
-image sets:
-  smashicons:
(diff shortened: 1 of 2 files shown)


BryceStevenWilley (approved):
LGTM; the only other place I see it is in https://github.com/A2JatAKCourts/docassemble-AKA2JBranding/blob/7a7b31484d5d5a4c3247f0cae00da096f976c058/docassemble/AKA2JBranding/data/questions/custom_organization.yml#L50, tagging @CaroRob just so she knows

PR #739, opened 2023-08-21
 Remove all file-like objects, not just ones at top level before saving answer sets

### LINKED ISSUE #680: When loading an answer set that had a signature stored in it, user is getting error "None object has no attribute seek"
Error is popping up when trying to display a template that has `skip undefined` turned off just before the signature screen should get triggered.

We do explicitly exclude any `file like` objects from the stored answer set on both loading and saving, but something may not be working correctly there.

nonprofittechy:
Fix #680


docassemble/AssemblyLine/sessions.py
```diff
@@ -1065,15 +1077,47 @@ def get_filtered_session_variables(
...
+    while items_to_check:
+        key, value = items_to_check.pop()
+
+        # This condition only will apply to "top level" variables
+        if is_file_like(value):
+            del all_vars[key]
+            continue
+
+        if isinstance(value, DAObject):
+            # docassemble overrides both __dir__ and __getattr__ for reasons
```
BryceStevenWilley, line 1095:
```suggestion
            # docassemble overrides both __dir__ and __getattr__ for reasons unknown
```

BryceStevenWilley (approved):
LGTM

PR #743, opened 2023-08-22
Better docstrings for `as_zip(title="", ...)`

BryceStevenWilley:
Say specifically where the zip title is used, improve a few PDF/A docstrings, and pass through pdfa param in a corner case.

Fixes #734, to the best that we can.


nonprofittechy (approved):

nonprofittechy:
@BryceStevenWilley only thing is I think you linked the wrong issue #. I can try to dig around for it but if you have it handy, can you link + close

PR #750, opened 2023-08-25
Inline terms of use

### LINKED ISSUE #629: Show terms of use in-line
It should also be possible to skip the terms of use if the user is logged in and has already accepted them.

Maybe we need a way to track which version/when they accepted the terms of use so that they can be re-prompted to accept the terms if they changed.

nonprofittechy:
Fix #629 

This adds 3 new nouns:

* AL_DEFAULT_CONTACT_INFORMATION
* AL_PRODUCTION_SITE_HOSTING_PROVIDER
* AL_TERMS_OF_USE_GOVERNING_JURISDICTION

Which are now documented https://suffolklitlab.org/docassemble-AssemblyLine-documentation/docs/framework/magic_variables#terms-of-use-specific-variables


docassemble/AssemblyLine/data/questions/al_terms_of_use.yml
```diff
@@ -0,0 +1,72 @@
+---
+code: |
+  # Customize this with the preferred contact information. It could be a physical address
+  # rather than an email.
+  AL_DEFAULT_CONTACT_INFORMATION = get_config("server administrator email", AL_ORGANIZATION_TITLE)
```
BryceStevenWilley, line 5: nit: IMO `AL_ORG_CONTACT_INFO` is a better name, it's not clear what it's the default for

docassemble/AssemblyLine/data/questions/al_terms_of_use.yml
```diff
@@ -0,0 +1,72 @@
...
+---
+code: |
+  AL_PRODUCTION_SITE_HOSTING_PROVIDER = AL_ORGANIZATION_TITLE
```
BryceStevenWilley, line 8: nit: I think `AL_HOSTING_PROVIDER` is succinct enough; I think `PRODUCTION_SITE` is implied / is a lil misleading (the hosting provider is also hosting the dev site likely, right? The terms could be dynamic to the server).

BryceStevenWilley (commented):
I don't know if I see all of this PR as useful; inline terms are a good additiob, but I don't see anyone using the terms template and changing the jurisdiction or hosting provider. It's a good template, I just don't see anyone using a dynamically updatable version from our library, and I wouldn't encourage them to.

I guess we do need a default, but I don't know if we need the 3 custom vars in that case.

The linked issue has a lot of other features that aren't in here; do we still want to do those and keep that issue open, or do we not think those are as useful?

nonprofittechy:
> I don't know if I see all of this PR as useful; inline terms are a good additiob, but I don't see anyone using the terms template and changing the jurisdiction or hosting provider. It's a good template, I just don't see anyone using a dynamically updatable version from our library, and I wouldn't encourage them to.
> 
> I guess we do need a default, but I don't know if we need the 3 custom vars in that case.
> 
> The linked issue has a lot of other features that aren't in here; do we still want to do those and keep that issue open, or do we not think those are as useful?

I opened a new issue to track those ideas. I think they are good ideas still, just hinge on the profile feature, but that's a bigger feature than this one.

I did also think at first we could do this without any dynamic variables, but I realized how likely it was that whatever we get from the others vars would be incomplete or wrong, or the logic would be messy to embed directly in the template.  E.g., jurisdiction is usually a state in the US, but many countries don't have states as a concept with its own laws. Same with contact information; the best I can default to is the email in the config, but we chose to use a physical address to discourage people using it for legal help requests. There's no config for physical address. Similarly, hosting provider is only sometimes going to be the same as the organization that made the interview.

I was thinking of specific interviews that I have worked on when I made all three new variables.

I think with these 3 variables many organizations will work out of the box, which is pretty handy, but we avoid making assumptions about the kind of choices people would make.

I could see this template being generated at a different level too though, perhaps in the Weaver interview we were imagining for generating an ALThemeTemplate. If we want to avoid extra variables I'm open to reconsidering.

nonprofittechy:
Thanks for suggestions, ultimately I'm reducing this to one new "magic variable" but it's an optional global configuration option to display the contact information as something other than an email address.

PR #763, opened 2023-09-12
Limit maximum height of logo

### LINKED ISSUE #748: We should limit the height of the logo to 40 pixels by default
We instruct authors that they can add their own limit, but the default, especially for SVG images, is not right with our theme.

This shouldn't be a breaking change as any organization that uses a custom logo likely already added our suggested CSS.

This is the CSS to add: 
```
.al-logo img {
  max-height: 40px;
}
```

nonprofittechy:
fix #748

This just mirrors the defaults in various production repos so more projects will work out of the box:

https://github.com/SuffolkLITLab/docassemble-ALWeaver/blob/fef677dd2daa9dd65d04b5621841f8e464910e86/docassemble/ALWeaver/data/static/styles.css#L25

https://github.com/SuffolkLITLab/docassemble-ALThemeTemplate/blob/2b0a0140a816c12f366a93feea1cdc7cdb95eb4e/docassemble/ALThemeTemplate/data/static/colors.css#L7

### DIFF
--- a/docassemble/AssemblyLine/data/static/styles.css
+++ b/docassemble/AssemblyLine/data/static/styles.css
@@ -12,7 +12,12 @@
   justify-items: start; 
   align-items: center; 
 }
-.al-logo { grid-area: logo; }
+.al-logo { 
+  grid-area: logo; 
+}
+.al-logo img {
+  max-height: 40px;
+}
 .al-title {
   display: grid; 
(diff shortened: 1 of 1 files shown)


BryceStevenWilley (approved):
LGTM, confirmed that it won't change ILAO's, and I agree with changing default for everyone else.

PR #766, opened 2023-09-23
Safer loading of YAML file with language names

nonprofittechy:
Corrected behavior of languages.py; it now will use a local languages.yml in the rare event that it exists, but correctly loads the package version of languages.yml in every other circumstances.

This was a breaking issue in the 209A but only visible on the test server so far.

### DIFF
--- a/docassemble/AssemblyLine/language.py
+++ b/docassemble/AssemblyLine/language.py
@@ -1,4 +1,5 @@
 # coding=utf-8
+import os
 from typing import List, Optional, Tuple
 from docassemble.base.util import url_action, path_and_mimetype
 import yaml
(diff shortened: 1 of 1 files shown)


docassemble/AssemblyLine/language.py
```diff
@@ -13,6 +14,32 @@
 ]
 
 
+def _package_name(package_name: Optional[str] = None):
+    """Get package name without the name of the current module"""
```
BryceStevenWilley, line 18:
```suggestion
    """Get package name without the name of the given module. By default this is `docassemble.AssemblyLine.language`"""
```

If I understand this correctly (that `__name__` should always be the `language.py` module). "Current module" is just vague enough that we should avoid it IMO

docassemble/AssemblyLine/language.py
```diff
@@ -13,6 +14,32 @@
...
+        0
+    ]
+    try:
+        local_yaml = path_and_mimetype("data/sources/languages.yml")[0]
+    except:
+        local_yaml = None
+
+    if local_yaml and os.path.isfile(local_yaml):
+        return local_yaml
+    return this_yaml
```
BryceStevenWilley, line 40:
```suggestion
    try:
        local_yaml = path_and_mimetype("data/sources/languages.yml")[0]
    except:
        local_yaml = None

    if local_yaml and os.path.isfile(local_yaml):
        return local_yaml
        
    al_package_name = _package_name()
    this_yaml = path_and_mimetype(f"{al_package_name}:data/sources/languages.yml")[
        0
    ]
    return this_yaml
```

Just moved `this_yaml` to the bottom (no need to create it if it's not going to be used), and renamed `current` to `al`, which makes things clearer IMO.

BryceStevenWilley (commented):
I couldn't really understand this at first go because I think think it's not clear what "current" vs "local" refers to. I think "al" vs "local" is much clearer, so I made so suggestions that might help. Otherwise, looks good

PR #769, opened 2023-09-24
Patch ALExhibit

### LINKED ISSUE #768: Recent DA version apparently breaks ALExhibit class
I know DAFileList had a recent change, assuming that causes this new error. (https://github.com/jhpyle/docassemble/releases/tag/v1.4.73)

```
Error

Interview has an error. There was a reference to a variable ‘exhibit_doc_defaults_1.exhibits[0].pages[1].initialized’ that could not be looked up in the question file (for language ‘en’) or in any of the files incorporated by reference into the question file.
```

Tried adding 

```
---
generic objeect: ALExhibitList
code: |
  x[i].pages[j].initialized = True
```

which didn't change behavior

nonprofittechy:
Fix #768 - should be backwards compatible

### DIFF
--- a/docassemble/AssemblyLine/data/questions/ql_baseline.yml
+++ b/docassemble/AssemblyLine/data/questions/ql_baseline.yml
@@ -1738,6 +1738,11 @@ validation code: |
 ---
 generic object: ALExhibitList
 code: |
+  # exhibit_doc_defaults_1.exhibits[0].pages[1].initialized
+  x[i].pages[j].initialized = True
+---
+generic object: ALExhibitList
+code: |
   x.there_are_any = x.has_exhibits
 ---
 generic object: ALExhibitList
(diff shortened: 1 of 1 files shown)


BryceStevenWilley:
Adding a comment from Jonathan from slack that helped me understand this change:

> The change in 1.4.73 gave DAFileList an object_type when previously it had no object_type.
When a list some_list has no object_type and you refer to an undefined some_list[1], then docassemble looks for a block that defines some_list[1]. But when there is an object_type, docassemble already knows how to define some_list[1]; it simply initializes a new object of that class.
Thus, after 1.4.73, your question that defines x[i].pages[j] (as it was written) will not be called upon.

Confirmed that it fixed, and also confirmed that it's backwards compatible (at least with 1.4.67, what we have on prod right now).

BryceStevenWilley (approved):

PR #776, opened 2023-10-04
Handle unexpected interview_metadata

BryceStevenWilley:
Ran into an error when `interview_metadata["main_interview_key"}` was set to a dictionary. Definitely not expected use for us, but it shouldn't error out if that's the case.

This PR adds a check to make sure the interview key is a string like we expect it to be. If it's not, we fallback to the other methods (just checking the only entry in `interview_metadata` like we do in the next check works in this case).

### DIFF
--- a/docassemble/AssemblyLine/data/questions/al_visual.yml
+++ b/docassemble/AssemblyLine/data/questions/al_visual.yml
@@ -356,7 +356,7 @@ content: |
   "${all_variables(special='metadata').get('title','').rstrip()}" version 
   `${ package_version_number }`; AssemblyLine version `${ al_version }`.
 
-  % if interview_metadata.get("main_interview_key"):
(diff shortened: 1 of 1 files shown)


nonprofittechy (commented):
Not a big fan of this kind of inline type checking to handle a developer error. I think it's just an error that should raise an exception, esp. since we are deprecating this variable. Am I missing the reason it was a dictionary in the interview you found in the wild?

BryceStevenWilley:
That's fair. Even if we're deprecating this variable, I think it'll be a while before this code can be removed.

> Am I missing the reason it was a dictionary in the interview you found in the wild?

It was in a MLH interview was just finishing up development. Given it was a simpler fix than #777 (which also makes me think there could be other metadata block issues), I figured it could be merged sooner, but we don't have to fix it.

PR #777, opened 2023-10-04
Move Metadata over to metadatablock

BryceStevenWilley:
In https://github.com/SuffolkLITLab/docassemble-ALWeaver/pull/835, we moved a lot of the metadata from the `interview_metadata` variable to the `metadata` block. This PR finishes the job, and allows AssemblyLine to look in the `metadata` block, if there isn't an `interview_metadata` at all. This works for newly weaved forms, and I've confirmed that it works great for interviews with and without `interview_metadata` as well.

The only tricky part was that `original_forms` in the metadata block has become a list of URLs, instead of just one. IMO that's fine, but I had to refactor some things and define one more mako variable to make the code clean. When there's only one original form, it'll still show as normal, and if there are more than one, it'll show raw URLs for each. Which is fine, not great, but I'm not sure if it's better to just list "original form 1", "original form 2", etc. Open to feedback here.

### DIFF
--- a/docassemble/AssemblyLine/data/questions/al_visual.yml
+++ b/docassemble/AssemblyLine/data/questions/al_visual.yml
@@ -366,10 +366,18 @@ content: |
       MAIN_METADATA = next(iter(interview_metadata.values())) 
     %>
   % else:
-    <% MAIN_METADATA = {} %>
+    <% MAIN_METADATA = all_variables(special='metadata') %>
   % endif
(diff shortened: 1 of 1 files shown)


nonprofittechy (commented):
This is looking good to me, but let me know when it's ready for a final review. Thanks for finishing this work

BryceStevenWilley:
Good timing, just tested on the Weaver and made sure existing tests pass!

nonprofittechy (approved):

PR #779, opened 2023-10-06
Load court list from correct package

BryceStevenWilley:
If you refer to plain file name without the path (i.e. "my_courts.xlsx") in an objects block, and then include that YAML from a different package, ALCourtLoader will fail to load the courts correctly, because it thinks that "my_courts.xlsx" should be in that different package, instead of in the original.

This fixes that by using
`docassemble.base.functions.this_thread.current_question.package`. Took the code from DAStaticFile._get_unqualified_reference, which is throughly used. Manually tested using https://github.com/mplp/docassemble-mlhframework.

### DIFF
--- a/docassemble/AssemblyLine/al_courts.py
+++ b/docassemble/AssemblyLine/al_courts.py
@@ -4,6 +4,7 @@ Package for a very simple / MVP list of courts that is mostly signature compatib
 import os
 from typing import Any, Dict, List, Optional, Union, Set
 import pandas as pd
+import docassemble.base.functions
 from docassemble.base.util import (
     path_and_mimetype,
     Address,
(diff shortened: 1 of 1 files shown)


BryceStevenWilley:
> I'm assuming you also considered and discarded the approach here, or a similar path manipulation approach?

I don't know what part of the file you're talking about there, there's a few functions that manipulate paths? `_package_name` seems like it always gets `docassemble.ALWeaver` how it's used, and that's the opposite of what we want.

> would it be simpler to use user_info().current_package?

`user_info().current_package` gets its value from `this_thread.current_question.from_source.package`. TBH I don't know how those might differ, or why they would differ. IMO the simplest thing is to do what DA does.

PR #782, opened 2023-10-11
More safety around phone numbers

BryceStevenWilley:
Always show the user given phone number if the formatted number is None.

Adds tests for this specific issue. Will merge when said tests pass.

### DIFF
--- a/docassemble/AssemblyLine/al_general.py
+++ b/docassemble/AssemblyLine/al_general.py
@@ -843,26 +843,24 @@ class ALIndividual(Individual):
         """
         nums = []
         if hasattr(self, "mobile_number") and self.mobile_number:
+            fmt_number = None
             try:
-                nums.append(
-                    {
-                        phone_number_formatted(
-                            self.mobile_number, country=country
-                        ): "cell"
-                    }
-                )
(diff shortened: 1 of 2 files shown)


nonprofittechy (approved):
This change looks good, but out of curiosity, in what circumstance was it failing before? I see there was already an existence and non-falsy value check for phone_number and mobile_number.

BryceStevenWilley:
Good question, I meant to put the relevant stack trace:

```
  File "/usr/share/docassemble/local3.10/lib/python3.10/site-packages/docassemble/AssemblyLine/al_general.py", line 828, in phone_numbers
    [
  File "/usr/share/docassemble/local3.10/lib/python3.10/site-packages/docassemble/AssemblyLine/al_general.py", line 829, in 
    list(num.keys())[0] + " (" + list(num.values())[0] + ")"
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
```
From the code I changed, I think `phone_number_formatted` from docassemble was returning `None` for some reason. It at least happens when people enter just letters, like this:

```python
from docassemble.base.util import phone_number_formatted
print(phone_number_formatted("aoeu"))
# shows "None"
```

And probably can happen at other points as well. It seems like we expected it to throw, but it never seemed to do that.

PR #783, opened 2023-10-12
Reduce children name questions to 1 question to avoid developer confu…

### LINKED ISSUE #781: Can we have 1 children[i].name_field() question?
It is difficult to understand and then overwrite these 2 so generic questions - I found it very confusing.
I didn't realize there are **2** questions that need to be used or overwritten. 
And the questions as they are in the code are difficult to understand. 
#780

plocket:
…sion about how to override questions. Closes #781

### DIFF
--- a/docassemble/AssemblyLine/data/questions/ql_baseline.yml
+++ b/docassemble/AssemblyLine/data/questions/ql_baseline.yml
@@ -816,30 +816,16 @@ validation code: |
 # this is just a very generic way to ask.
 id: children names without birthdate
 sets:
-  - children[0].name.first
-  - children[0].name.last
-  - children[0].name.middle
-  - children[0].name.suffix
-question: |
-  % if hasattr(children, 'ask_number') and children.ask_number and children.target_number == 1:
-  Child's name
-  % else:
-  Name of first child
-  % endif
-fields:
-  - code: |
-      children[0].name_fields()
----
-# Customize this question if you know whose children they are in your interview!
-# this is just a very generic way to ask.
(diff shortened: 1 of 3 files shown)


docassemble/AssemblyLine/data/questions/ql_baseline.yml
```diff
@@ -815,31 +815,17 @@ validation code: |
...
-# Customize this question if you know whose children they are in your interview!
-# this is just a very generic way to ask.
-id: children names without birthdate
 sets:
   - children[i].name.first
   - children[i].name.last
   - children[i].name.middle
   - children[i].name.suffix
 question: |
+  % if hasattr(children, 'ask_number') and children.ask_number and children.target_number == 1:
```
nonprofittechy, line 824:
```suggestion
  % if hasattr(children, 'ask_number') and children.ask_number and children.target_number == 1 and i == 0:
```

plocket: I'm happy to make the change, but I just want to check on my understanding. If the target number is 1, won't `i == 0` always resolve to `True`?

nonprofittechy: no. target number is only used when you use .gather(). If there's an "add" button you can get past it.

nonprofittechy (approved):
This looks fine. I don't totally agree it's easier to have one question than 2 for customizing purposes, but I think this change doesn't make it a lot harder either.

PR #784, opened 2023-10-13
Include AssemblyLine through interview list

BryceStevenWilley:
At MLH, we make a `interview_list` wrapper recently, that just included our interview framework YAML and `interview_list.yml`.

However, in doing so, `al_audio.js` is added twice to the page twice, and runs twice, so there are two Voice RSS widgets that show up, and one of them doesn't properly work (since it has the same id as the first).

IMO, `interview_list.yml` should just use AssemblyLine fully, instead of the awkward half-includes that it currently does. The only changes that happen when you do so are the red `pre` debug text and the footer, both of which I manually set to empty. People are welcome to re-enable those on `interview_list` if they wish.

I'm working an a PR to stop the JS stuff from running twice, even if it's on the page twice, still in progress there.

Downstream issue: https://github.com/mplp/docassemble-mlhframework/issues/15

### DIFF
--- a/docassemble/AssemblyLine/data/questions/interview_list.yml
+++ b/docassemble/AssemblyLine/data/questions/interview_list.yml
@@ -2,19 +2,14 @@
 # Provide a replacement for the /interviews endpoint in Docassemble
 ---
 include: 
-  - docassemble.ALToolbox:display_template.yml
-  - al_settings.yml
+  - assembly_line.yml
 ---
 features:
   small screen navigation: dropdown
   navigation: True
   navigation back button: False
-  javascript:
-    - al_audio.js
   css:
-    - styles.css
-    - al_audio.css
-    - interview_list.css    
+    - interview_list.css
   question help button: True
   question back button: False
 ---
@@ -79,6 +74,9 @@ default screen parts:
         <span class="title-row-2">${ all_variables(special='metadata').get('short title','').rstrip() }</span>
       </span>
     </span>
+  # These two were originally blank before including all of AssemblyLine.
+  pre: ""
+  footer: ""
 ---
 mandatory: True
 code: |


nonprofittechy (commented):
It might make sense to make some epics for some changes to Assembly Line so we can plan the changes and I can describe the things that need to be tested before it can be merged [before you start designing a solution so it's not just my feedback on a PR]

This fix might work. But when someone uses the normal styled assembly line page, you get our default navigation bar title, etc. which are clearly placeholders. I do know that I don't want to have a lot of default branding on servers that use just the interview list. A key design goal is that someone shouldn't have to customize this page or any server settings to have a default that is a drop-in replacement for the normal interview list page. 

I'm pretty sure I tried something like your approach first and it broke for some real-world use cases (including Pleading Power, who used the interview list before Suffolk did). Maybe the only issue was the double audio input.

Not ready to merge this without identifying what needs to be tested a bit more thoroughly and having a chance to revisit my original reasoning.

BryceStevenWilley:
> A key design goal is that someone shouldn't have to customize this page or any server settings to have a default that is a drop-in replacement for the normal interview list page.

I'm okay with that, I didn't realize that was the case, given it's in AssemblyLine and not ALToolbox, and given how it's implemented now. When `interview_list.yml` loads, it defines / gets the default definitions of:

* `AL_ORGANIZATION_HOMEPAGE` (the "Start a new form" button leads to it)
* `AL_ORGANIZATION_TITLE` (still set in the default parts of `interview_list.yml`
* `al_logo` (also still set in default parts)

Agreed that an epic for stuff like this would be a good idea. #785 fixes the downstream issue independently as well, so we can it you want.

> I can describe the things that need to be tested before it can be merged [before you start designing a solution so it's not just my feedback on a PR]

Sorry, I do kinda shoot off the hip with PRs sometimes. This one wasn't really designing a solution as much as it was diagnosing that it was the issue in the first place, and most things that seem like bugs and not design decisions I will make quick PRs for.

nonprofittechy:
> > A key design goal is that someone shouldn't have to customize this page or any server settings to have a default that is a drop-in replacement for the normal interview list page.
> 
> I'm okay with that, I didn't realize that was the case, given it's in AssemblyLine and not ALToolbox, and given how it's implemented now. When `interview_list.yml` loads, it defines / gets the default definitions of:
> 
>     * `AL_ORGANIZATION_HOMEPAGE` (the "Start a new form" button leads to it)
> 
>     * `AL_ORGANIZATION_TITLE` (still set in the default parts of `interview_list.yml`
> 
>     * `al_logo` (also still set in default parts)

Ah, I guess it does more than I thought. I couldn't think of a way around the `AL_ORGANIZATION_HOMEPAGE` but it does default to /list. We could probably get rid of the logo and title.

> > I can describe the things that need to be tested before it can be merged [before you start designing a solution so it's not just my feedback on a PR]
> 
> Sorry, I do kinda shoot off the hip with PRs sometimes. This one wasn't really designing a solution as much as it was diagnosing that it was the issue in the first place, and most things that seem like bugs and not design decisions I will make quick PRs for.

No worries there--if we check-in before big stuff that's good enough for me. I just feel bad shooting down an idea when it's already in a PR.

PR #785, opened 2023-10-13
Don't add our al_audio if existing node is hidden

BryceStevenWilley:
Otherwise, we end up adding our controls twice, sometimes with the same id and everything, so the second isn't displayed right.

Additionally, I added the event param to the `daPageLoad` callback function; when I was trying to debug this, being able to use the browser JS debugger was very useful, but I couldn't edit the running code at all, so having more event information to look at would have helped a bunch.

* Downstream issue: https://github.com/mplp/docassemble-mlhframework/issues/15
* Related AssemblyLine issue / alternate fix: https://github.com/SuffolkLITLab/docassemble-AssemblyLine/pull/784

### DIFF
--- a/docassemble/AssemblyLine/data/static/al_audio.js
+++ b/docassemble/AssemblyLine/data/static/al_audio.js
@@ -1,7 +1,7 @@
 // Do not overwrite already created al_js object. This may be relevant as our js gets built out.
 if ( !al_js ) { var al_js = {}; }
 
-$(document).on('daPageLoad', function() {
+$(document).on('daPageLoad', function(event) {
   try {
     var $audio_nodes = $('.daaudio-control');
     var id_count = 1;
@@ -10,7 +10,7 @@ $(document).on('daPageLoad', function() {
       al_js.replace_with_audio_minimal_controls( audio_node, 'page_reader_substitute_' + id_count );
       id_count++;
     }
-  } catch ( error ) { console.log( 'AL audio instantiation error:', error ) }
+  } catch ( error ) { console.log( 'AL audio instantiation error:', error, 'event was ', event ) }
 });
 
 // We are not providing a way to rewind or scan through the audio.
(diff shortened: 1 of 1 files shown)


nonprofittechy (approved):

plocket:
Just wanted to mention that this looks pretty good, but a bit indirect - checking the da audio node and assuming what we have or haven't done about that. I think a more robust solution might be to just check whether the id exists and return if it does. I love the extra info in the logs!

I'm going to test and then review 👍 

Unrelatedly, I also think `al_` at the beginning of the id would be great.

plocket (commented):
Worked for me! I gave my caveat in my comment and I'm fine with whatever you folks decide.

BryceStevenWilley:
Thanks for the thoughts, I was waiting for them before I merged!

> Just wanted to mention that this looks pretty good, but a bit indirect - checking the da audio node and assuming what we have or haven't done about that.

I had the same reservations at first, but thinking through it more, we shouldn't add controls to a hidden element anyway, since when we do, the new element won't be hidden. Adding coding to try to sync the two elements and un-hide ours when the original element is unhidden would be difficult and error-prone.

> I think a more robust solution might be to just check whether the id exists and return if it does. I love the extra info in the logs

I actually tried that at first; explicitly keeping track of which elements we've added our controls to. But the Voice RSS elements that DA puts on the page don't have ids, so we can't really tell them apart.

I see what you mean about not creating elements with duplicate ids. That's a good idea, but IMO it doesn't necessarily mean that we shouldn't add our controls to the element, just that it needs a different id. The ids we add aren't intrinsically related to the element, just it's order on the page. Making up hyptotheticals robustness scenarios: if DA switched to loading the VoiceRSS element dynamically, then the first element that shows up in our search for `.daaudio-control` wouldn't have been there at page load, and we'd still need to add our control, even if we had already added a `page_reader__1` id earlier.

I agree this implementation has some downsides, but IMO it's the most robust out of the options.

Issue #281, opened 2021-08-28
Make address fields dynamic, based on the selected country for the address field, handle international addresses


nonprofittechy:
MVP: update the "states list" if someone selects a country outside of the US. Also update the label of "zip code" to "postal code"

Beyond MVP: figure out which countries we can have country-specific address fields for, and which countries might benefit instead from just an open text input. We could probably just hardcode a small list of fields for countries that our client population frequently encounters, such as Canada and Mexico, although we might also consider using all of the fields that Google knows about and using a Python library that helps figure out which fields are relevant in a particular country. See https://docassemble.org/docs/objects.html#Address

```
    myaddress.street_number - the street number (e.g., 123).
    myaddress.street - the street name (e.g., Main St).
    myaddress.city - the city (known as (locality).
    myaddress.county - the county (known as administrative_area_level_2).
    myaddress.state - the state (known as administrative_area_level_1).
    myaddress.zip - the Zip code (known as postal_code).
    myaddress.country - the country (e.g., US) in ISO 3166-1 alpha-2 format.
    myaddress.administrative_area_level_1
    myaddress.administrative_area_level_2
    myaddress.administrative_area_level_3
    myaddress.administrative_area_level_4
    myaddress.administrative_area_level_5
    myaddress.colloquial_area
    myaddress.floor
    myaddress.intersection
    myaddress.locality
    myaddress.neighborhood
    myaddress.post_box
    myaddress.postal_code
    myaddress.postal_code_prefix
    myaddress.postal_code_suffix
    myaddress.postal_town
    myaddress.premise
    myaddress.room
    myaddress.route
    myaddress.sublocality
    myaddress.sublocality_level_1
    myaddress.sublocality_level_2
    myaddress.sublocality_level_3
    myaddress.sublocality_level_4
    myaddress.sublocality_level_5
    myaddress.subpremise
```

See #127, and also the way that the "state" field is made dynamic when someone selects a country in https://github.com/SuffolkLITLab/docassemble-ALWeaver/pull/478/files

nonprofittechy:
Some code related to this that I'm removing from its original branch for now: 

```
        # if show_country:
        #     state_field = {
        #         "label": (f"% if not defined('{ self.attr_name('country') }'):\n" +
        #                   str(self.state_label) + "\n" +
        #                   f"% elif safe_subdivision_type({ self.attr_name('country')}):\n" +
[...]

PR #794, opened 2023-11-24
International addresses

BryceStevenWilley:
Minor changes that address #281. More details in that issue, but I don't think it's possible to get the full state dropdown for each country without messing up Google autocomplete, so when the country is an input, we turn the state list into a text box (I tried swapping out a dropdown for just the US and text input for other countries, but Google autocomplete also has a problem with that).

Only other changes were to change the "State" label to "State / Province" when the country is an input, and a few comments / docs, for accuracy.

Also relevant to https://github.com/mplp/docassemble-MLHMotionRegardingChildSupport/issues/6.

### DIFF
--- a/docassemble/AssemblyLine/al_general.py
+++ b/docassemble/AssemblyLine/al_general.py
@@ -176,17 +176,6 @@ class ALAddress(Address):
             ]
         else:
             fields = []
-        if show_country:
-            fields.append(
-                {
-                    "label": str(self.country_label),
-                    "field": self.attr_name("country"),
-                    "required": False,
-                    "code": "countries_list()",
-                    "default": country_code,
-                }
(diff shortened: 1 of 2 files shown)


nonprofittechy:
Smart and simple. This is a nice enhancement, thanks! Lgtm

BryceStevenWilley:
Minor changes that address #281. More details in that issue, but I don't think it's possible to get the full state dropdown for each country without messing up Google autocomplete, so when the country is an input, we turn the state list into a text box (I tried swapping out a dropdown for just the US and text input for other countries, but Google autocomplete also has a problem with that).

Only other changes were to change the "State" label to "State / Province" when the country is an input, and a few comments / docs, for accuracy.

PR #803, opened 2023-12-11
Check SMS is enabled before showing texting opts

### LINKED ISSUE #411: Make some features (signature, sharing) that depend on multi_user or a valid email/sms config responsive to whether it's turned on or off
I think we might want some Python functions to check for a valid SMS and email configuration.

I think we'll need to audit the default interview text to look for places we offer the ability to send by email or SMS and conditionally turn those off. In most places we allow both methods, and in some there's a backup method of clicking a link to copy text, so we probably don't want to turn these totally off in the footer.

BryceStevenWilley:
* `is_sms_enabled()`, sees if the `twilio` config in the settings are non-empty (doesn't check that they work)
* Turns off mentions of texting in the "Share" page
* Turns off option to get texts from the reminder system
* Turns off options to text the link to yourself when signing

The only change I could see being suggested is to move `is_sms_enabled()` (and `is_phone_or_email()`, even thought it's been in AL for a while) to the Toolbox. IMO, it's not worth splitting highly coupled functions between two different packages, but will happily do it if suggested.

Fixes #411 for texting, not for email, but as mentioned in the meeting this morning, most of our users have some sort of email enabled on their server.

### DIFF
--- a/docassemble/AssemblyLine/data/questions/al_visual.yml
+++ b/docassemble/AssemblyLine/data/questions/al_visual.yml
@@ -195,14 +195,20 @@ fields:
     datatype: radio
     code: |
       sharing_choices
-  -  How do you want to share the link?: al_how_share_link
-     datatype: radio
-     choices:
-       - Email${ " or text message" if is_sms_enabled() else ""}: email_or_sms
-       - Just show me the link. I will share it myself.: link_only
-       - Export my answers to a file: download_json
-     js show if: |
-       val("al_sharing_type") === "tell_friend" || val("al_sharing_type") === "share_answers" 
+  - How do you want to share the link?: al_how_share_link
+    datatype: radio
+    choices:
+      - label: |
+          % if is_sms_enabled():
+          Email or text message
+          % else:
+          Email
+          % endif
+        value: email_or_sms
+      - Just show me the link. I will share it myself.: link_only
+      - Export my answers to a file: download_json
+    js show if: |
+      val("al_sharing_type") === "tell_friend" || val("al_sharing_type") === "share_answers"
   - note: |
       **Note**: the person you share this link with will be able to see and
       edit your answers on this form.
@@ -219,7 +225,13 @@ fields:
       ${ copy_button_html( interview_url(temporary=48), label=al_copy_button_label.show(), tooltip_inert_text=al_copy_button_tooltip_inert_text.show(), tooltip_copied_text = al_copy_button_tooltip_copied_text.show()) }
     js show if: |
       val("al_sharing_type") === "share_answers" && val("al_how_share_link") === "link_only"
-  - Email ${ "or phone number " if is_sms_enabled() else ""}you want to send this to: share_interview_contact_method
+  - label: |
+      % if is_sms_enabled():
+      Email or phone number you want to send this to
+      % else:
+      Email you want to send this to
+      % endif
+    field: share_interview_contact_method
     validate: is_phone_or_email
     show if:
       variable: al_how_share_link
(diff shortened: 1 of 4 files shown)


docassemble/AssemblyLine/data/questions/al_visual.yml
```diff
@@ -193,13 +193,12 @@ question: |
...
-    choices:
-      - Tell a friend about this website: tell_friend
-      - Share my answers and progress with someone: share_answers
+    code: |
+      sharing_choices
   -  How do you want to share the link?: al_how_share_link
      datatype: radio
      choices:
-       - Email or text message: email_or_sms
+       - Email${ " or text message" if is_sms_enabled() else ""}: email_or_sms
```
nonprofittechy, line 201:
```suggestion
       - label: |
           % if is_sms_enabled():
           Email or text message
           % else:
           Email
           % endif
         field: email_or_sms
```

Let's avoid ternary statements for user-facing text--languages change word order, and this kind of expression is going to be mis-read by the translator. See: https://suffolklitlab.org/docassemble-AssemblyLine-documentation/docs/coding_style_guide/yaml_translation#avoid-using-the-language-functions-from-patternen-in-your-interview

docassemble/AssemblyLine/data/questions/al_visual.yml
```diff
@@ -220,7 +219,7 @@ fields:
       ${ copy_button_html( interview_url(temporary=48), label=al_copy_button_label.show(), tooltip_inert_text=al_copy_button_tooltip_inert_text.show(), tooltip_copied_text = al_copy_button_tooltip_copied_text.show()) }
     js show if: |
       val("al_sharing_type") === "share_answers" && val("al_how_share_link") === "link_only"
-  - Email or phone number you want to send this to: share_interview_contact_method
+  - Email ${ "or phone number " if is_sms_enabled() else ""}you want to send this to: share_interview_contact_method
```
nonprofittechy, line 222: See comment above about translation

docassemble/AssemblyLine/data/questions/al_visual.yml
```diff
@@ -311,7 +317,7 @@ template: al_share_answers_message_template
 subject: |
   ${ AL_ORGANIZATION_TITLE } form from ${ al_share_form_from_name }
 content: |
-  ${ share_interview_answers_message }
+  ${ share_interview_answers_message.replace("\n", "<br>") }
```
nonprofittechy, line 320: Are we always sending HTML email?

nonprofittechy (changes requested):
Looking good--I really like this change. Can't merge until we handle language properly. And maybe a clarification about the sending of HTML email. If we do make that change, we should use proper syntax - `<br/>` not `<br>`

docassemble/AssemblyLine/data/questions/al_visual.yml
BryceStevenWilley: Yeah, docassemble always sends an HTML version. Relevant code is: https://github.com/jhpyle/docassemble/blob/70d7dbfd7db5075f1e51d2b64906739a8f86f679/docassemble_base/docassemble/base/util.py#L8366-L8371

Specifically, when you pass in a template, it gets the content of that template in HTML. BeautifulSoup strips out the tags for the plaintext version if the client doesn't support HTML (which most email clients support, and default to).

BryceStevenWilley:
> If we do make that change, we should use proper syntax - `<br/>` not `<br>`

I'm fine with changing it, but from what I understand, `<br>` is fine, `<br/>` is only necessary for XHTML

nonprofittechy (approved):

PR #824, opened 2024-02-14
Add proxy variables' HTML for ALKiln >=5.9.0 2-column Story Table

### LINKED ISSUE #823: Update ALKiln `post:` to add new 2-column Story Table HTML element
Code to put in `post:`

```mako
    <div id="alkiln_proxy_var_values"
    data-generic_object="${ encode_name(str( x.instanceName if defined('x') else '' )) }"
    % for letter in [ 'i', 'j', 'k', 'l', 'm', 'n' ]:
    data-index_var_${ letter }="${ encode_name(str( value( letter ) if defined( letter ) else '' )) }"
    % endfor
    aria-hidden="true" style="display: none;"></div>
```

plocket:
Proxy variables are generic objects and index variables. I don't think this needs to be added anywhere else.

Closes #823

### DIFF
--- a/docassemble/AssemblyLine/data/questions/al_code.yml
+++ b/docassemble/AssemblyLine/data/questions/al_code.yml
@@ -186,8 +186,18 @@ code: |
     # If the user role was set to "unknown" at the time wizard run
     user_role = user_ask_role
 ---
-# A bit of HTML required for ALKiln to be used in interviews.
-# To add to your unstyled alkiln interviews, include it in your `default screen parts: post`
+# Old HTML for ALKiln deprecated 3-column Story Tables.
 template: alkiln_trigger_html
 content: |
-    <div data-variable="${ encode_name(str( user_info().variable )) }" id="trigger" aria-hidden="true" style="display: none;"></div>
+  <div data-variable="${ encode_name(str( user_info().variable )) }" id="trigger" aria-hidden="true" style="display: none;"></div>
+---
+# HTML for interviews using ALKiln tests at ALKiln versions 5.9.0 and above.
+# If you are using al_package_unstyled.yml, include it in your `default screen parts: post`
+template: alkiln_proxy_html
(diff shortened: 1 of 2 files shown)


docassemble/AssemblyLine/data/questions/al_code.yml
```diff
@@ -186,8 +186,18 @@ code: |
...
     user_role = user_ask_role
 ---
-# A bit of HTML required for ALKiln to be used in interviews.
-# To add to your unstyled alkiln interviews, include it in your `default screen parts: post`
+# Old HTML for ALKiln deprecated 3-column Story Tables.
 template: alkiln_trigger_html
 content: |
-    <div data-variable="${ encode_name(str( user_info().variable )) }" id="trigger" aria-hidden="true" style="display: none;"></div>
\ No newline at end of file
+  <div data-variable="${ encode_name(str( user_info().variable )) }" id="trigger" aria-hidden="true" style="display: none;"></div>
```
nonprofittechy, line 192: Could we reasonably combine this with the template above, or are people likely to want one but not both?

plocket: It's a good question. I want people to use the new one, not both. There's no reason for new tests to use the old one. Assembly Line probably needs to keep it around for now to keep people's old tests working, but I want new Story Tables to avoid it if possible.

I also I think keeping that old id ("trigger") around with the new stuff would only add to confusion as it's not relevant anymore. It's hard enough in both our internal and external documentation to explain what's going on with Story Tables - decoding and guessing variable names, etc.

That said, if you feel it'll be detrimental to AssemblyLine, ALKiln can either do as you describe or make an exception to handle interviews that do so.

nonprofittechy (approved):
I'll trust your judgment on the separate templates. Please feel free to merge once tests are passing

PR #837, opened 2024-03-25
When using the as_pdf() method of a bundle, add parameter to ensure an odd or even number of pages

### LINKED ISSUE #828: Make it possible to ensure that a bundle has an even number of pages
This would modify the `as_pdf()` method so that it would pdf_concatenate all docs, and then add an extra, blank PDF only if the page count was odd.

The purpose of this is to help with double-sided printing. An author would have to create sub-bundles for each doc that needs to have an even page count.

Alternatively, we could also add this parameter to the ALDocument.as_pdf() method.

This is only necessary for docs that can be expanded (DOCX files)

nonprofittechy:
Adds a parameter `ensure_parity` to the `ALDocumentBundle.as_pdf()`, and `default_parity` as an attribute of the bundle.

parity means odd or even number of pages. When either parameter or attribute is set, the as_pdf() method of the bundle may pad the output with an additional page, if needed to match the desired parity.

The purpose is primarily to assist with double-sided printing.

- `ALDocumentBundle.default_parity`: can be "odd", "even", or `None`.
- `ALDocumentBundle.as_pdf(ensure_parity)` can be "odd", "even", or `None`. Overrides `default_parity`.

Test document: `test_bundle_parity.yml`.

Example usage:

```
---
objects:
  - even_bundle: ALDocumentBundle.using(
          title="Even bundle",
          filename="the_bundle",
          elements=[
            the_doc
          ],
          default_parity="even",
          enabled=True
      )
```

Fix #828

### DIFF
--- a/docassemble/AssemblyLine/al_document.py
+++ b/docassemble/AssemblyLine/al_document.py
@@ -1,7 +1,7 @@
 import re
 import os
 import mimetypes
-from typing import Any, Dict, List, Union, Callable, Optional
+from typing import Any, Dict, List, Literal, Union, Callable, Optional
 from docassemble.base.util import (
     Address,
     LatitudeLongitude,
@@ -34,6 +34,7 @@ from textwrap import wrap
 from math import floor
 import subprocess
 from collections import ChainMap
+import pikepdf
 
 __all__ = [
     "ALAddendumField",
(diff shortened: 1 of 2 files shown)


docassemble/AssemblyLine/al_document.py
```diff
@@ -180,6 +181,50 @@ def table_row(title: str, button_htmls: List[str] = []) -> str:
     return html
 
 
+def pdf_page_parity(pdf_path: str) -> Literal["even", "odd"]:
+    """
+    Count the number of pages in the PDF and
+    return "even" if it is divisible by 2 and "odd"
+    if it is even.
```
samglover, line 188: I think this should read "if it is odd."

samglover (commented):
Spotted one typo in the documentation. Setting aside punctuation errors in the code, which I'm not yet confident I would spot, it seems like it should do what it is supposed to!

docassemble/AssemblyLine/al_document.py
nonprofittechy, line 188:
```suggestion
    return "even" if it is divisible by 2 and "odd"
    if it is not divisible by 2.
```

PR #841, opened 2024-04-25
Pad .zip attribute with 0 and ensure proper loading from court list

### LINKED ISSUE #840: Ensure address_zip of a court list is treated as a string with leading zeros
See: https://github.com/LSVermont/docassemble-VTSharedYMLFile/issues/9

nonprofittechy:
Fix #840

Currently, when you use the court_loader class in al_courts.py, there's a chance that a column that has a zip code with leading zeros will be interpreted incorrectly.

This fixes it in 2 ways:

1. transforms the `address_zip` column inside the dataframe when it is loaded to zerofill it to at least 5 digits
2. when printing a zip code (.zip attribute) as part of an address, zerofill to 5 digits.

It also adds an API to define a transformation for arbitrary columns of the Excel spreadsheet: defining the `.converters` attribute of the `ALCourtLoader` will now automatically run the converters when the dataframe is loaded. This honors the standard Pandas syntax, where `converters` should be a mapping between the column name and a callable with one parameter.

Ex:

```
---
objects:
  - all_courts: |
      ALCourtLoader.using(file_name='courts_list.xlsx', converters={"address_zip": float})
```

### DIFF
--- a/docassemble/AssemblyLine/al_courts.py
+++ b/docassemble/AssemblyLine/al_courts.py
@@ -435,15 +435,17 @@ class ALCourtLoader(DAObject):
         def convert_zip(z: Any) -> str:
             return str(z).zfill(5)
 
-        merged_converters: Dict[str, Callable[[object], object]] = {"address_zip": convert_zip}
+        merged_converters: Dict[str, Callable[[object], object]] = {
+            "address_zip": convert_zip
+        }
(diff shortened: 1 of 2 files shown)


samglover:
The logic of this fix makes sense and the ALKiln error doesn't seem to relate to it.

PR #846, opened 2024-04-25
Allow pronoun functions to work when .pronouns is defined as a string rather than DADict

### LINKED ISSUE #845: Let the pronouns functions work even if the pronouns attribute is a string,
Right now we expect the pronouns to be a dictionary like this:

```python
{"he/him/his": True,
 "she/her/hers": False}
```

but if the developer made this radio buttons instead so it's a string like `they/them/theirs`, we should still be able to parse that.

nonprofittechy:
Fix #845


samglover (approved):
This makes sense to me!

PR #849, opened 2024-04-30
Handle False boolean overflow triggers, not just True

nonprofittechy:
It's possible to set a `boolean` as the `overflow_trigger` for an ALAddendumField when the conditions for triggering the addendum don't depend on the length of a list or string. But we weren't properly handling the condition where the `boolean` was `False`.

This handles the `False` condition in the overflow trigger by returning an empty string, which will result in `has_overflow()` returning `False`.

This situation may be a bit unusual, but it might be worth using this pattern when the conditional text is required to be added to the same addendum as other overflow text. In other situations, it would make more sense to create a standard ALDocument with the `enabled` value set.

This also fixes a production bug in the Vermont Divorce package.


samglover (approved):
Looks good!

PR #852, opened 2024-05-13
Add an optional `maxlength` attribute to any `_fields` methods

nonprofittechy:
@mnewsted I'm opening this up as a draft PR for now, LMK if the API makes sense.

### DIFF
--- a/docassemble/AssemblyLine/al_document.py
+++ b/docassemble/AssemblyLine/al_document.py
@@ -776,13 +776,13 @@ class ALAddendumFieldDict(DAOrderedDict):
         self[the_key].field_name = the_key
         return newobj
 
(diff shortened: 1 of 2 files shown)


mnewsted:
Thanks @nonprofittechy ! I appreciate you adding this to a host of _fields methods.

If I understand correctly, will just add the maxlength=# attribute inside a function call, like users[i].name_fields(person_or_business='unsure', maxlength=40) ? And it applies to all internal fields, in this case (first, middle, last, and business name?

nonprofittechy:
It would look like this:

users[0].name_fields(max length={"first": 40, "last": 32})

mnewsted:
That's even better. The ability to customize length on a field-by-field basis = chef's kiss.

PR #861, opened 2024-06-12
Make it possible to disambiguate 'familiar' names

### LINKED ISSUE #857: Should .familiar() default to showing the suffix in addition to first name?
This would disambiguate situations where there are is a parent and a child on the same form.

E.g., John JR or April IV

nonprofittechy:
Fix #857

Returns the individual's name in the most familiar form possible.

The purpose is to allow using a short version of the individual's name in an unambiguous
way in the interview. For example: referring to the child in a guardianship petition
by first name instead of "the minor". But there may be a problem if context doesn't make
it clear if you are talking about the child or their parent when they share a name.

In order, it will try to use:

* just the first name
* the first name and suffix
 * the first and middle name
 * the first and last name
 * the full name
 * the default value, e.g., "the minor", if provided
 * the full name

Example usage:
```mako
Who do you want to take care of ${ children.familiar(unique_names=parents + petitioners, default="the minor") }?
```

### DIFF
--- a/docassemble/AssemblyLine/al_general.py
+++ b/docassemble/AssemblyLine/al_general.py
@@ -1797,7 +1797,7 @@ class ALIndividual(Individual):
 
     def familiar(
         self, unique_names: Optional[List[Any]] = None, default: Optional[str] = None
-    ):
+    ) -> str:
         """
         Returns the individual's name in the most familiar form possible.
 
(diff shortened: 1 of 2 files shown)


ekressmiller:
This seems like a good idea. I hadn't thought of that pitfall of using just the first name but in retrospect it's so clear!

PR #864, opened 2024-06-18
Add `find_matching_sessions` function to enable keyword searching across sessions, based on metadata

nonprofittechy:
Adds a new function, `find_matching_sessions`, which can be used to search and retrieve a set of matching sessions across all unencrypted interview sessions on a server.

The expected use case is for an attorney or other advocate to be able to search through their own sessions to find a document they automated for a particular user. It can also be used in a scenario where there is a generic "intake" interview that the attorney wants to use as a starting point to automate future documents with. Search results can be limited to a particular interview or set of interviews, such as an intake questionnaire. 

Searching across sessions that you did not create or were not explicitly shared is limited to 3 user roles by default: admin, developer, and advocate.

Matching is performed entirely based on keywords in the `jsonstorage` table; by default, only ones with the metadata key name set to "metadata." If your interview uses the Assembly Line, it automatically saves the name of the `users` and `other_parties` lists to the `jsonstorage` table with this key on each screen load.

Example:

```python
matching_sessions=find_matching_sessions("smith", user_id="all", filenames=[f"{user_info().package}:intake.yml", "docassemble.MyPackage:intake.yml"])
```

### DIFF
--- a/docassemble/AssemblyLine/data/questions/demo_search_sessions.yml
+++ b/docassemble/AssemblyLine/data/questions/demo_search_sessions.yml
@@ -2,6 +2,14 @@
 include:
   - assembly_line.yml
 ---
+metadata:
+  title: |
+    Search sessions demo
+  required privileges:
+    - admin
+    - developer
+    - advocate
+---
 mandatory: True
 code: |
   search_keyword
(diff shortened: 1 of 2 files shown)


samglover (approved):
Looks good to me!

PR #878, opened 2024-08-15
Add support for reflexive pronouns

nonprofittechy:
This adds a new method, pronoun_reflexive(), to both ALPeopleList and ALIndividual.

A reflexive pronoun is a word like "myself", "yourself", "themselves", "ourselves", "herself", or "himself" that acts upon itself.

When used on a singular person with pronouns they/them or gender "nonbinary", it will return "themself", which is increasingly being used as a standard singular pronoun for a person who is nonbinary or uses they/them pronouns.

When used on a list of multiple people, it will return "themselves".

the pronoun_reflexive() pronoun also respects 1st and 2nd person point of view using the Docassemble standard methods to indicate point of view, including passing the "person" parameter.

### DIFF
--- a/docassemble/AssemblyLine/al_general.py
+++ b/docassemble/AssemblyLine/al_general.py
@@ -804,6 +804,44 @@ class ALPeopleList(DAList):
             and_string=and_string,
         )
 
+    def pronoun_reflexive(self, **kwargs) -> str:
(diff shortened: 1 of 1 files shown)


samglover (approved):
Good change. LGTM!

PR #880, opened 2024-09-09
Split purpose of github_repo_name and github_user from dual purposes; more detailed comments

nonprofittechy:
Help code detectives who miss the current explanation in https://assemblyline.suffolklitlab.org/docs/framework/magic_variables#github_repo_name and https://assemblyline.suffolklitlab.org/docs/framework/magic_variables#github_user

Also splits the functional meaning of 2 separate concepts in a backwards compatible way


samglover (approved):
Makes a lot of sense after our problem-solving last week.

PR #890, opened 2024-10-28
Basic background assembly implementation

### LINKED ISSUE #184: Allow downloading a single PDF of the combined bundle, possibly on the same row that the ZIP download option will appear

### LINKED ISSUE #788: Create a version of download_list_html that shows pre-assembled documents from a background task
See #84 #219 

One quick solution for larger document sets is to use a background task to assemble documents. `download_list_html` currently assumes the documents are generated live to prevent folks from getting "stale" versions of the documents.

I think a quick way to help with document assembly problems is just to create some background tasks and a way to reconsider the task on the download screen that can be used manually for the interviews that we think need it now. We can still keep working on more robust/improved background assembly methods over time.

nonprofittechy:
Fix #788 fix #184 

This refactors the document generation functionality in download_list_html into a new function, `get_cacheable_documents`, and adds some simple events that can be used for background assembly.

Additionally, it adds a single downloadable PDF row to the download screen.

This adds 2 tasks that are available for any bundle:

`generate_downloads_with_docx_task` and `generate_downloads_task`. Each task populates the `_downloadable_files` attribute of the bundle, either with or without the accompanying original DOCX files (when available).

`test_aldocument_background_assembly.yml` shows an example workflow for using the new background assembly functionality.

@KelizaMcGuire


samglover (approved):
Not sure I would spot a code problem, but it sounds like a good upgrade.

samglover (approved):

ekressmiller:
would it be possible to add the use_previously_cached_files option to send_email as well so that it could likewise generate the needed files in the background?

ekressmiller:
@nonprofittechy, is there a way to use this method while still allowing an "Edit" button (taking you to the review screen) on the download screen?

I just tested:
1. arrive at the download screen
2. click button to go to the review screen and change a variable
3. resume interview, which returns me to the download screen

Since use_previously_cached_files is True, the change I made was not reflected in the assembled docs.

nonprofittechy:
no, it doesn't work with the edit button. I suggest instructing the user to hit the back button, and making the previous screen a review screen that always appears

PR #907, opened 2025-02-18
Add a backwards-compatible /s/ signature option (it makes a picture)

### LINKED ISSUE #661: Add support for /s/ signature method
We realized that digital signature field isn't very accessible

* /s/Some Name is a valid digital signature in most fields
* We think digital signatures get edited in the PDF by Docassemble (maybe?)

We should make a new top level variable name to match the way we use the "sign on paper" option

### LINKED ISSUE #858: Sign on a phone with finger is not working
This bug has been experienced by an independent tester, Jack H. and me.

I was on my laptop and used phone to open QR code. On my phone it brought up a signature screen that didn't have a box to sign in. Had to hit Undo on phone to go back to signature choice screen. Then could move forward and sign.

Also, for second option on screen we could say "On a computer or other device with my mouse, finger or trackpad." Couldn't we? On my touchscreen laptop, I was able to use my finger or trackpad.
 
image

 - docassemble 1.4.104
 - AssemblyLine version AL-2.28.1.

### LINKED ISSUE #891: Allow for mobile user to choose on-paper signature
When you use a phone, currently it doesn't allow you the choice of signing online or not signing at this time. You have to sign online or you can't move on. It would be great to allow interview authors the ability to change this setting. Thanks!

### LINKED ISSUE #896: Have a standard way to indicate a /s/ signature
We have learned from MA Trial Court that some courts are okay with a /s/ signature but do not accept a picture of a signature, for whatever reason. /s/ signature is also a bit easier for the user to perform.

Two options:

1. We illustrate the user's signature, by generating an image that can be inserted into the existing field. This is the most backwards compatible method and just needs an update to the `signature_if_final()` method.
2. We use the text of the user's name as text, but we add it to a standard field with a standardized name.

For example, we can add a field named like `users1_signature_text` with a transparent background and place it in the exact spot, but on top of the digital signature field.

I will be using this second method on at least a couple of forms.

nonprofittechy:
This also fixes a bug affecting some interviews, where the user could not continue if they used a QR code to get a screen to sign on their phone. AFAICT, the issue was with a Docassemble interview that had `note` fields but no visible fields to capture an answer because they were hidden with `show if`. The bug must have been introduced in the last 6 months.

And it adds an option for mobile users to sign in a way other than with a drawn signature (sign after printing or sign with /s/)

Fix #896, fix #891, fix #858, fix #661 

See also https://github.com/SuffolkLITLab/docassemble-MATCSmallClaims/issues/127

The /s/ signature option is always available. Authors would need to override the signature choice screen to remove this option. However, in most jurisdictions that we've worked in, /s/ is the preferred digital signature option.

This feature adds two new configuration options that can control the digital signature's appearance:

1. `al_typed_signature_prefix`: defaults to "/s/" if not specified
2. `al_typed_signature_font`: defaults to Google's ["Bad Script" font](https://fonts.google.com/specimen/Bad+Script) if not specified.

Because the signature is created as an image, this is completely backwards compatible with a standard, drawn signature. It will be usable in a PDF without editing the field type. Here is an example in the default Bad Script:

image

Note the Kiln failure is not related to this PR


samglover (approved):
LGTM!

Issue #903, opened 2025-02-10
Add custom label parameter to send_button_html() method


mnewsted:
Could you add a parameter to the .send_button_html() method for ALDocumentBundle so the button label can be customized? It is currently using a font-awesome envelope icon and "Send." That works great most of the time, though we might want to customize.

Please let me know if you need more information.

Thanks!

Issue #902, opened 2025-01-14
Allow developers to adjust non-button label of zip file row in .download_list_html()


mnewsted:
When generating the download table of an ALDocumentBundle, I would like to be able to adjust the text that appears in the final row in the left column (where the entire bundle can be downloaded of a .zip file). The .download_list_html() function used to use the ALDocumentBundle "title," though it now uses zip_label. Thanks to @plocket for tracking this down!

I hope there's a way to restore the ability to customize the text. Using the .title attribute of the ALDocumentBundle would be ideal since that's how our code is written. Maybe it could use the zip_label if no title is defined? Let me know how I can help or if you have any questions. Thanks!

PR #909, opened 2025-03-03
Add docs custom send button label & icon optional keyword args

plocket:


plocket:
Sorry, conflated #902 with this one, #903. Will have to work on #902 another day.

PR #910, opened 2025-03-11
Switch back to using bundle title for zip row label. Add override keyword argument.

### LINKED ISSUE #902: Allow developers to adjust non-button label of zip file row in .download_list_html()

plocket:
Closes #902

I restored the old behavior of using `.title` as the row label (the text in the left-hand column). It seemed reasonable since [the bundle's individual documents use their `title` attribute for their row label](https://github.com/SuffolkLITLab/docassemble-AssemblyLine/blob/2b28ad2b11aad154071b2b754e183a490e5d3ab3/docassemble/AssemblyLine/al_document.py#L2030). I wasn't completely sure what you meant in our messages, @nonprofittechy, so I added an explicit keyword argument that can be used as an override to that value.

I'm not sure the last fallback to the zip row label value [assignment] is necessary.


plocket:
Working on the type checking error.

plocket:
Commit message was supposed to be "Use falsey value", but you can't always get what you want

plocket:
Doing my best, but apparently not good enough. Just the typing failures.

nonprofittechy (approved):

PR #917, opened 2025-03-27
Allow more debug operations on production environment

### LINKED ISSUE #915: Add a custom URL to turn the red `pre` area on (sometimes called `HUD`)
E.g., `debug=1`.

Maybe should only take effect if the user is a privileged user - developer, admin, or perhaps advocate, but maybe safe for all environments?

nonprofittechy:
Fix #915

Adds a URL parameter, `debug=1`, that allows a developer, admin or advocate to:

1. View the standard HUD
2. Quickly get to the /vars?i=... URL for the running interview file.

These should both help with debugging errors in the production environment more easily.

Note that these parameters aren't available for a regular user.

### DIFF
--- a/docassemble/AssemblyLine/data/questions/al_visual.yml
+++ b/docassemble/AssemblyLine/data/questions/al_visual.yml
@@ -41,11 +41,14 @@ default screen parts:
   corner back button label: |
     Undo
   pre: |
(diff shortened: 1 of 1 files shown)


samglover (approved):
LGTM

PR #919, opened 2025-04-18
Support dynamically provided options in the DOCX template file

nonprofittechy:
If you are using a DOCX template, this adds a new Jinja2 filter, `catchall_options(value [implicit], *items)`, that lets you define the options in the template itself. The options will be presented as a drop down menu in the interview.

Template options would typically be provided with a syntax like either of the below examples:

```
    {{ my_catchall_field | catchall_options("code1: label1", "code2: label2") }}
    {{ my_catchall_field_2 | catchall_options({"code1": "label1"}, {"code2": "label2"}) }}
```

This is useful for situations when you have a very simple interview that you would like to allow an end-user to edit and upload and you have DACatchAll enabled to flag undefined variables. E.g., some legal aid programs use this to request client signatures.

This also adds a filter `catchall_label` with similar functionality, but this allows you to define the `label` of a catchall question in the template.

Together, these new functions allow much of the functionality of the interview to be driven by the template author, requiring less work inside the interview YAML. This is helpful for a lot of kinds of templates that are used by attorneys rather than end-users.

Note that this special syntax only works with DACatchall objects. You cannot use it, for example, to override the options if there is an existing question in the YAML with the same variable name.

### DIFF
--- a/docassemble/AssemblyLine/custom_jinja_filters.py
+++ b/docassemble/AssemblyLine/custom_jinja_filters.py
@@ -86,4 +86,34 @@ def catchall_options(value: Any, *raw_items: Any) -> DACatchAll:
     return value
 
 
+def catchall_label(value: Any, label:str) -> DACatchAll:
+    """Jinja2 filter to allow you to define a label for a DACatchAll field inside a DOCX template.
+
+    This filter takes a label string and assigns it to the `label` attribute of the
+    DACatchAll object. This label can be used to provide a more descriptive name for the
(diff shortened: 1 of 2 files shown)


nonprofittechy:
@BryceStevenWilley no rush, but this probably does deserve a review by someone more familiar with the internals of Docassemble.

Also @ll-jgraszer

samglover (approved):
Excellent feature. LGTM!

PR #930, opened 2025-06-30
Check for person type before showing last name

### LINKED ISSUE #929: If someone enters a person name and then changes it to a business, last name is still shown
We could check for the person_type attribute and not show the last name if it's a business

nonprofittechy:
fix #929


BryceStevenWilley (approved):
Added some minimal unit tests, but LGTM!

PR #934, opened 2025-07-08
Add an optional arg to require zip / postal code

BryceStevenWilley:
Necessary for E-filing, as Tyler requires a Zip code be provided when registering an account. When users don't enter the zip code (as it's not indicated as required), they see an error.


BryceStevenWilley:
> a parameter required that has a dictionary with the required/optional state that can override the default required state.

To make sure I'm understanding, something like this? `address_fields(..., required = {"unit": False, "zip": True})`

Issue #773, opened 2023-09-28
Stop updating the metadata multiple times per page load


BryceStevenWilley:
Since https://github.com/SuffolkLITLab/docassemble-AssemblyLine/pull/747, I've noticed that the the block that updates metadata happens multiple times on the same page load. This routinely lengthens page load times to 0.3 or 0.5 seconds. By itself it's not much, but it is a small irk, does make page load times longer when they don't need to be.

We should check if there's a way to only run `initial` blocks once per screen load. It's not clear why it's running multiple times IMO. 

```yaml
initial: True
code: |
  ########### Save title and progress to session list ##################
  if get_config("assembly line",{}).get("update session metadata", True):
    al_temp_metadata = get_interview_metadata(
      user_info().filename, user_info().session
    )
    al_temp_metadata["auto_title"] = str(al_sessions_interview_title)
    al_temp_metadata["progress"] = get_progress()
    set_current_session_metadata(al_temp_metadata)
    del(al_temp_metadata)
```

<details><summary>Full DA stack dump</summary>
<pre>
How question came to be asked
Ran initial code at 0.00000s

(from al_language.yml)

initial: True
code: |
  if enable_al_language:
    set_language(al_user_language)

Ran initial code at 0.00005s

(from al_saved_sessions.yml)

initial: True
code: |
  ########### Save title and progress to session list ##################
  if get_config("assembly line",{}).get("update session metadata", True):
    al_temp_metadata = get_interview_metadata(
      user_info().filename, user_info().session
    )
    al_temp_metadata["auto_title"] = str(al_sessions_interview_title)
    al_temp_metadata["progress"] = get_progress()
    set_current_session_metadata(al_temp_metadata)
    del(al_temp_metadata)

Needed definition of al_sessions_interview_title at 0.00096s
Tried to run block at 0.00110s

(from al_saved_sessions.yml)

template: al_sessions_interview_title
content: |
  % if defined('users[0].name.first'):
  ${ comma_and_list(users.complete_elements()) }
  % if defined('other_parties[0].name.first'):
  v. ${ comma_and_list(other_parties.complete_elements()) }
  % endif
  % else:
  ${ all_variables(special='titles').get('full') }
  % if all_variables(special='titles').get('sub'):
  : ${ all_variables(special='titles').get('sub') }
  % endif
  % endif

Ran initial code at 0.00127s

(from al_language.yml)

initial: True
code: |
  if enable_al_language:
[...]

PR #935, opened 2025-07-11
Slightly more efficient and simplified metadata update method

nonprofittechy:
This doesn't seem to have a huge impact, but it might shave about 0.02s from a typical page load?

I tested this on the test_question_library.yml file and that's the difference I was seeing.

Completed processing at 0.15401s the old way, vs 0.13143s the new way. I probably didn't do enough passes to be positive, but I like the API a bit better anyway. It's not likely to be a huge impact either way.

This is part of the solution to #773 . Want to test @jpagh's idea also. But this seems safe and helpful.

### DIFF
--- a/docassemble/AssemblyLine/custom_jinja_filters.py
+++ b/docassemble/AssemblyLine/custom_jinja_filters.py
@@ -119,5 +119,126 @@ def catchall_label(value: Any, label: str) -> DACatchAll:
     return value
 
 
+def catchall_subquestion(value: Any, subquestion: str) -> DACatchAll:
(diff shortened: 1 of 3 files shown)


BryceStevenWilley (approved):
LGTM! Not the most knowledgable about Postgres Jsonb, but it checks out from what I can tell.

I don't think that the speed up is really worth it, but I think having a better API for updating session metadata is good, same with getting rid of the extra tmp var we needed to delete in interviews.

Agree that a slight race condition is fine for this feature, session metadata doesn't need to be exactly in sync with on every page.

PR #941, opened 2025-07-21
Use responsive language in user contact methods question

### LINKED ISSUE #918: Contact methods screen - specific to users[0] but should it be?
We have a unique "how to contact me" for users[0], but when someone customizes this screen, that means they need to customize both users[0] and users[i] version of the screen.

for a lot of other questions, we've already consolidated and replaced with conditional logic based on the value of `i`.

Would want to check with others using our framework, especially @mnewsted , to see if this change would be disruptive.

nonprofittechy:
Fix #918 

Also updated the translation_es.xlsx file with this change.


BryceStevenWilley (approved):
LGTM

PR #942, opened 2025-07-21
Use `al_international_phone` datatype for all phone number fields

### LINKED ISSUE #552: Use the  `datatype: al_international_phone` for phone numbers as the default input type

nonprofittechy:
This shouldn't affect e-filing screens as they are already customized.

Fix #552


BryceStevenWilley (approved):
LGTM!

PR #943, opened 2025-07-21
Allow customizing pronoun list and disabling the neopronoun ze/zir/zirs both case-by-case and globally

### LINKED ISSUE #927: Let someone configure which pronouns are shown using pronoun_fields
My team @lemmalegalconsulting works on a lot of estate planning interviews, and disabling the ze/zir/zirs pronouns is a frequent request. I think we had strong reasons for including this as a default in our nonprofit, public facing tools, but a simple way for folks to disable these choices would be helpful for other situations.

Would also be helpful to let someone add their own neopronouns.

nonprofittechy:
Adds a new parameter to the ALIndividual.pronoun_fields() method, `choices`, which allows overriding the default list of choices for `pronoun_fields()`

If the parameter is not provided, we will use a magic variable (from the interview) `al_pronoun_choices`.

The most common expected use case is to remove `ze/zir/zirs` from the list.

(discussion below reflects previous version of this description)

Fix #927

### DIFF
--- a/docassemble/AssemblyLine/data/questions/test_question_library.yml
+++ b/docassemble/AssemblyLine/data/questions/test_question_library.yml
@@ -40,7 +40,7 @@ subquestion: |
   % elif item == "pronouns":
   Your pronouns are: ${ users[0].list_pronouns() }.
 
-  In the past, ${users[0].pronoun_subjective(person="2") } called ${ users[0].pronoun_reflexive(person="2") } ${ users[0].pronoun_objective(person="1") }.
+  In the past, ${users[0].pronoun_subjective(person="3") } called ${ users[0].pronoun_reflexive(person="3") } ${ users[0].pronoun_possessive('', person="3") }.
   % else:
   ${ getattr(users[0], item) }
   % endif
--- a/docassemble/AssemblyLine/data/questions/test_question_library.yml
+++ b/docassemble/AssemblyLine/data/questions/test_question_library.yml
@@ -17,18 +17,41 @@ code: |
 
   if which_to_test == 'address':
     users[0].address.address
-  if which_to_test == 'address_no_address':
+  elif which_to_test == 'address_no_address':
     users[0].address.has_no_address
-  if which_to_test == 'contact':
+  elif which_to_test == 'contact':
     users[0].phone_number
+  elif which_to_test == 'pronouns':
+    users[0].pronouns
   all_done
 ---
 event: all_done
 question: All done!
+subquestion: |
+  % for item in ["name", "pronouns", "address", "address_no_address", "contact"]:
+  % if which_to_test == item:
+  % if item == "contact":
+  Your contact information is:
+  ${ users[0].email }[BR]
+  ${ users[0].phone_numbers() }[BR]
+  ${ users[0].other_contact_method}[BR]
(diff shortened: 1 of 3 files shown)


nonprofittechy:
@BryceStevenWilley @samglover feedback is welcome as to whether adding a global configuration option for this makes sense. I can say that the `pronoun_fields()` methods are widely used in the estate planning interviews I've helped with, and those do not use default AssemblyLine screens, so it feels like adding a new `al_use_ze_pronouns` magic variable wouldn't be very helpful for folks who do want to change the default behavior of this method.

samglover:
I guess the question is whether ze should be included by default, and if so whether to make it easy to remove from the defaults?

If ze has become popular enough to be included among the defaults, it makes sense to include a way to remove it, as well. Or if it isn't included among the defaults but is the next-most-popular pronoun, it should be easy to add it.

nonprofittechy:
> I guess the question is whether ze should be included by default, and if so whether to make it easy to remove from the defaults?
> 
> If ze has become popular enough to be included among the defaults, it makes sense to include a way to remove it, as well. Or if it isn't included among the defaults but is the next-most-popular pronoun, it should be easy to add it.

When we first added the `pronoun_fields()` method, we added `ze/zir/zirs`, and it's currently on by default.

The question in this PR is only whether adding a new global configuration option, which is pretty rare, is the right way for us to change the default behavior and disable using `ze/zir/zirs`

BryceStevenWilley:
I feel like my main opposition to this way of thinking is that there should be a specific param to exclude ze. I like `choicees` as a param, and I think whatever we do should have that as the main (i.e. only) param. So, a global config would be fine with me, but it should be something like:

```
assembly line:
  pronouns choices:
      - she/her/hers: She/her/hers
      - ze/zir/zirs: Ze/zir/zirs
      ...
```

etc. Puts the complexity onus on interviews that want to override our defaults, and not in our implementation.

Global configs do run into problems when we host interviews with multiple stakeholders who want different configs for their interviews. Slightly less of an issue now that we're in the server hosting business, but still might happen with us with appeals vs trial court vs just our stuff. But in general I'm not too opposed to it, it just almost always runs into limits.

docassemble/AssemblyLine/data/questions/test_question_library.yml
```diff
@@ -17,18 +17,41 @@ code: |
...
+  Your contact information is:
+  ${ users[0].email }[BR]
+  ${ users[0].phone_numbers() }[BR]
+  ${ users[0].other_contact_method}[BR]
+  % elif item == "address_no_address":
+  ${ users[0].address_block() }
+  % elif item == "pronouns":
+  Your pronouns are: ${ users[0].list_pronouns() }.
+
+  In the past, ${users[0].pronoun_subjective(person="2") } called ${ users[0].pronoun_reflexive(person="2") } ${ users[0].pronoun_objective(person="1") }.
```
BryceStevenWilley, line 43: When I tried this, it says "In the past, you called yourself me.", which I don't think is what you intended? The below shows the actual pronouns you selected.

```suggestion
  In the past, ${users[0].pronoun_subjective(person="3") } called ${ users[0].pronoun_reflexive(person="3") } ${ users[0].pronoun_possessive('', person="3") }.
```

PR #945, opened 2025-07-21
Improve custom error action screen - formatting, default open, and avoid chance for user to accidentally delete answers

### LINKED ISSUE #770: Make the error message fully expanded (or at least more details expanded) when the server is in debug mode
See: https://github.com/SuffolkLITLab/ALKiln/issues/763 for motivation

### LINKED ISSUE #797: Avoid buttons for exiting the interview
From the VT accessibility audit:

> Some < button > elements should really be links since their function is navigation; for example, “Go to
legal aid website” if the user can’t continue the interview. We recommend changing all
elements having a navigation purpose to links.

Probably isn't anything to change in this repo, but we definitely do the same in our interviews, and probably have recommended people do the same in the docs. Making this as an umbrella issue for those other fixes, but happy to transfer it / move it around.

### LINKED ISSUE #914: On the "error" screen, remove the exit button
In supporting staff people, sometimes they click "exit" and it's impossible to see what the problem was.

### LINKED ISSUE #944: Formatting is often broken on the custom error action
@IllinoisLegalAidOnline has fixed this with the simple change to using `indent()` instead of using `pre` for the stack trace.

Should copy that fix here, upstream.

nonprofittechy:
Fix #914 - use new_session instead of restart, fix #797
fix #944 formatting on error screen
fix #770, expand error screen on debug

image

image

# Without change

image

### DIFF
--- a/docassemble/AssemblyLine/data/questions/al_visual.yml
+++ b/docassemble/AssemblyLine/data/questions/al_visual.yml
@@ -527,13 +527,13 @@ subquestion: |
   % endfor
 
   % if not get_config("verbose error messages") == False:
-  ${ collapse_template(al_formatted_error_message) }
+  ${ collapse_template(al_formatted_error_message, collapsed=(not get_config("debug"))) }
   % endif
 back button label: Back
 buttons:
   - Exit: leave
     url: ${ AL_ORGANIZATION_HOMEPAGE }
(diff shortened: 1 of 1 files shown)


BryceStevenWilley (approved):
LGTM!

PR #946, opened 2025-07-22
allow most _fields() methods to accept passing a list of options

### LINKED ISSUE #657: Allow passing a function name to the _fields() methods - for name_suffix(), states_list(), etc.
We can improve flexibility of our `name_fields()`, `address_fields()` etc. function by allowing a form author to pass a lambda in as a parameter to customize the various lists that we use.

E.g., people sometimes want to customize the list of suffixes.

nonprofittechy:
Fix #657 

For `name_fields()` specifically, added a way to customize the list of options for both suffixes and titles globally without using a `language` function. 

Added two `data` blocks (so they are available in standard XLSX translation file) with two new global magic variables: `al_name_suffixes` and `al_name_titles`.

This is likely not considered backwards compatible because it ignores the definition of `name_suffix()` and the complicated language function translation system you can use with it. A simple `code` block in an interview can restore the old behavior, but the new extended list meets the needs of some of our community partners without removing any options and is translatable in a more standardized way.

When passed as parameters, you can use a `Callable` rather than a bare list. although this flexibility might not be important in most cases, I think it could help if you still want to use the standard `language` function method on an individual method-call basis.

### DIFF
--- a/docassemble/AssemblyLine/al_general.py
+++ b/docassemble/AssemblyLine/al_general.py
@@ -1,5 +1,5 @@
 from copy import deepcopy
-from typing import Dict, List, Literal, Union, Optional, Any
+from typing import Callable, Dict, List, Literal, Union, Optional, Any
 from docassemble.base.util import (
     Address,
     as_datetime,
@@ -21,7 +21,6 @@ from docassemble.base.util import (
     Individual,
     IndividualName,
     its,
-    name_suffix,
     phone_number_formatted,
     phone_number_is_valid,
     showifdef,
@@ -32,6 +31,7 @@ from docassemble.base.util import (
     this_thread,
     url_action,
     validation_error,
+    value,
     word,
     your,
 )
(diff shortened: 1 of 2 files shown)


docassemble/AssemblyLine/al_general.py
```diff
@@ -1084,11 +1087,13 @@ def name_fields(
                 Default is True.
             show_title: (bool, optional): Determines if the name's title (e.g., Mr., Ms.) should be included in the prompts.
                 Default is False.
-            title_options (List[str], optional): A list of title options to use in the prompts. Default is defined as a list
-                of common titles in English-speaking countries.
+            title_choices (Union[List[str], Callable], optional): A list or callable of title options to use in the prompts. Default is defined as a list
+                of common titles in English-speaking countries, or overridden by value of global `al_name_suffixes`.
```
nonprofittechy, line 1091:
```suggestion
                of common titles in English-speaking countries, or overridden by value of global `al_name_titles`.
```

PR #956, opened 2025-07-31
Don't explicitly close connection, instead use context manager

### LINKED ISSUE #955: InterfaceError: (psycopg2.InterfaceError) cursor already closed
I've seen this on other interviews before, but @ekressmiller has reported it happening with the `interview_list.yml` often, so I think it's worth tracking and looking into more.

```

[InterfaceErrorExample.txt](https://github.com/user-attachments/files/21517855/InterfaceErrorExample.txt)

nonprofittechy:
Fix #955 

Refactored both get_interview_metadata() and get_saved_interview_list() which were, by way of Docassemble's `variables_snapshot_connection()`, making use of raw psycopg2 connections instead of the higher-level sqlalchemy connections. Made better use of context manager which should also avoid premature session closing.

I don't recall a reason to use the variables_snapshot_connection() in these two functions, as the rest of the module uses the Docassemble shared `db` object.

`get_interview_metadata()` is used on the al_saved_sessions_store.yml page, and get_saved_interview_list() is used on the interview_list.yml page for basic testing. Testing the search feature particularly should validate this change.

### DIFF
--- a/docassemble/AssemblyLine/sessions.py
+++ b/docassemble/AssemblyLine/sessions.py
@@ -50,6 +50,7 @@ import os
 import re
 import hashlib
 import struct
+
 try:
     import zoneinfo  # type: ignore
 except ImportError:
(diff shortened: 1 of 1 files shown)


BryceStevenWilley (approved):
LGTM! You could probably remove the (no unused) import for `variables_snapshot_connection`, but could get cleaned up later too.

PR #961, opened 2025-08-18
Fix SyntaxWarning starting in 3.12

BryceStevenWilley:
This warning has started showing up when installing AssemblyLine on Python 3.12 and later:

```
SyntaxWarning: invalid escape sequence '\S'
  if re.match("\S+@\S+", text) or (sms_enabled and phone_number_is_valid(text)):
```

It's because the regex string isn't a raw string, so it's acting as escape sequences on `S` instead of the regex characters `\S`. The intended behavior is kept whether or not we have the raw string, but the warning goes away if we use the raw string.


nonprofittechy (approved):
Thanks for fixing this. I noticed the warning but hadn't had a chance to track down the solution yet!

PR #966, opened 2025-09-02
New Jinja2 filter, `if_final` that controls whether to show a placeholder or allow Docassemble to get the definition of a variable

nonprofittechy:
Justification:

When you render an ALDocument, you may want to have certain placeholder values that are shown on the preview screen (most commonly, a placeholder value for the signature) without triggering the definition of those variables.

This adds a new filter, `if_final` whose behavior depends on the value of `i`:

When `i` is "preview" (or any value other than `final`), the placeholder value is shown. The default placeholder is `[ variable or final attribute name in variable's name path ]`.

For example: when the undefined variable is `users[0].signature`, the placeholder becomes `[ signature ]`.

* The value of `i` is retrieved from the context of the template by default, although it can be passed explicitly as well.
* the `expected_values` can be either a string or a list of strings that should result in the true value instead of a placeholder.
* the `placeholder` can also be customized to a value other than the variable's name in brackets.

This new filter:

* Generalizes the behavior of the special method, `signature_if_final(i)` that only handles signatures of an ALIndividual
* Is a shorter and simpler syntax
* Should cover all of the common use cases for such a filter. I don't think a more generalized filter (e.g., handling the inverse situation of only applying when a certain value of `i` is not met) has any current uses, so I am focusing only on the useful filter for now.


jpagh:
LGTM! I think generalizing it out like this is the best option, and actually better than your original question.

docassemble/AssemblyLine/custom_jinja_filters.py
```diff
@@ -200,11 +203,140 @@ def catchall_subquestion(value: Any, subquestion: str) -> DACatchAll:
...
+    # 3) As a last resort, use a generic label
+    return str(placeholder)
+
+
+@pass_context
+def if_final(
+    context: Jinja2Context,
+    value: Any,
+    i: Optional[str] = None,
+    expected_values: Union[str, List[str]] = "final",
```
BryceStevenWilley, line 241:
```suggestion
    expected_i: Union[str, List[str]] = "final",
```

If we're hard coding this to just work with i, I'd change this variable name. It took me a while to realize it was talking about the values of `i`, and not the value that you're passing in `users[0].signature`, etc.

docassemble/AssemblyLine/custom_jinja_filters.py
```diff
@@ -200,11 +203,140 @@ def catchall_subquestion(value: Any, subquestion: str) -> DACatchAll:
...
 
+def _undefined_label(ud: Undefined, placeholder: Optional[str] = None) -> str:
+    """
+    Extract a friendly name from a Jinja2 Undefined.
+    Prefer the last missing attribute/key; fall back to the variable name;
+    then try parsing the hint; finally a generic label.
+
+    Args:
+        ud (Undefined): The Jinja2 Undefined object.
+        placeholder (str, optional): A generic label to use if no other name can be found.
```
BryceStevenWilley, line 214: Since `placeholder` isn't ever passed in, should it be removed, and the fallback just be "undefined"?

docassemble/AssemblyLine/custom_jinja_filters.py
```diff
@@ -200,11 +203,140 @@ def catchall_subquestion(value: Any, subquestion: str) -> DACatchAll:
...
+        str: A friendly name extracted from the Undefined object.
+    """
+    # 1) Best case: attribute/key name (e.g., "signature" from users[0].signature)
+    name = getattr(ud, "_undefined_name", None)
+    if isinstance(name, str) and name:
+        # Note that name is already going to be just the last part of a dotted variable name,
+        # whic is perfect for a descriptive-ish placeholder.
+        return name
+
+    # 2) Sometimes the hint contains 'foo' or "foo"
```
BryceStevenWilley, line 225: this comment was confusing, since I don't really know what the hint even is. Saying that we're looking for a string is better IMO.


```suggestion
    # 2) Sometimes the hint contains a string in quotes, like 'foo' or "foo"
```

docassemble/AssemblyLine/custom_jinja_filters.py
```diff
@@ -200,11 +203,140 @@ def catchall_subquestion(value: Any, subquestion: str) -> DACatchAll:
...
+    if (
+        i != expected_values
+        if isinstance(expected_values, str)
+        else i not in expected_values
+    ):
+        if isinstance(value, Undefined):
+            if not placeholder:
+                _label = _undefined_label(value)
+                if "." in _label:
+                    _label = _label.split(".")[-1]
```
BryceStevenWilley, line 325: When is there a dot in the label names? The docstring of `_undefined_label` specifically says that there shouldn't be a dot in the undefined name, so this logic being here is kinda confusing.

BryceStevenWilley (approved):
LGTM. Some suggestions to clean up unused params, but feature looks good!

docassemble/AssemblyLine/custom_jinja_filters.py
nonprofittechy: Fair enough! I really can't think of another solid use case that isn't to make an easier syntax for our special case with `i` but if we come up with one, it should have a different name.

PR #979, opened 2025-10-06
sr-only is no longer supported, so use visually-hidden instead. it seems forwards compatible

nonprofittechy:


BryceStevenWilley:
> I asked ChatGPT about the accessibility implications, if you're curious: https://chatgpt.com/share/68e3e10a-8b10-800d-97c5-cacecf6b8988

For what it's worth, most of that is wrong: `sr-only` and `visually-hidden` have the exact same implementation, `visually-hidden` just adds an `!important` to each CSS attribute.

https://github.com/twbs/bootstrap/blob/v5.0.0/dist/css/bootstrap.css#L6243

https://github.com/twbs/bootstrap/blob/v4.6.2/dist/css/bootstrap.css#L8101

samglover:
> > I asked ChatGPT about the accessibility implications, if you're curious: https://chatgpt.com/share/68e3e10a-8b10-800d-97c5-cacecf6b8988
> 
> For what it's worth, most of that is wrong …

LOL figures. Thanks for digging up the details and satisfying my curiosity!

PR #987, opened 2025-11-14
Fix bug where running an action, like visiting review page, unsets user language

nonprofittechy:
https://docassemble.org/docs/functions.html#url_action


BryceStevenWilley (approved):
LGTM!

PR #988, opened 2025-11-14
Support DALazyTemplate in ALAddendumField

### LINKED ISSUE #986: ALAddendum should handle template blocks like strings
This lets you combine multiple bits of text for a longer explanation without needing to use `depends on`.

Also make sure to add some examples.

nonprofittechy:
Fix #986

This helps when you have a few fields that you want to insert into a single box in a PDF. By putting each field into a `template` block, you can respect the final line/character count in a single field without complex calculations. By using a DALazyTemplate, the calculations will be done fresh for "free" without requiring you to use a `depends on` modifier or `reconsider: True`.

### DIFF
--- a/docassemble/AssemblyLine/al_document.py
+++ b/docassemble/AssemblyLine/al_document.py
@@ -156,10 +156,7 @@ def safeattr(object: Any, key: str) -> str:
                     object.location, LatitudeLongitude
                 ):
                     return ""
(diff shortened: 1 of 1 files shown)


BryceStevenWilley (approved):
I'd remove the unnecessary changes, but good otherwise.

PR #996, opened 2025-12-01
Return unique results based on the session ID

### LINKED ISSUE #888: Don't show the same session twice - use a unique constraint on the session ID when searching sessions

nonprofittechy:
Previously, it was possible to get duplicate results from find_matching_sessions and get_saved_interview_list. It seems this mostly would be the case when an advocate joined a session created by a user from the general public.

This is a modest fix that just adds SELECT DISTINCT to those queries.

Fix #888

### DIFF
--- a/docassemble/AssemblyLine/sessions.py
+++ b/docassemble/AssemblyLine/sessions.py
@@ -379,7 +379,7 @@ def get_saved_interview_list(
     filenames_to_exclude.extend([current_filename, filename_to_exclude])
 
     query_draft = """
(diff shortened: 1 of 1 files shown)


BryceStevenWilley:
FWIW, this is erroring on dev:

```
ProgrammingError: (psycopg2.errors.InvalidColumnReference) SELECT DISTINCT ON expressions must match initial ORDER BY expressions
LINE 2:         SELECT DISTINCT ON (userdict.key) userdict.indexno
                                    ^                     
 ```

PR #1004, opened 2026-01-20
Add bates number params to AL Document

BryceStevenWilley:
Lets the ALDocumentBundle use `bates_number` function when generating a PDF, as well as the work-in-progress params.

Depends on https://github.com/jhpyle/docassemble/pull/898

### DIFF
--- a/docassemble/AssemblyLine/al_document.py
+++ b/docassemble/AssemblyLine/al_document.py
@@ -2768,9 +2768,14 @@ class ALExhibit(DAObject):
         self,
         *,
         refresh: bool = False,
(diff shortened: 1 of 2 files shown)


nonprofittechy (approved):
LGTM!

nonprofittechy:
Changing to draft--update state when it's ready to merge, and we can do a major version bump if we need if this depends on a specific Docassemble release

BryceStevenWilley:
Forgot to comment back here, but fixed the remaining issues that I had found with the ALDocument (mainly being able to run with previously created objects. Should be ready for a re-review.

PR #1009, opened 2026-01-29
fix: resolved import error of `this_thread` by changing the import to come from `functions` instead of `util` due to docassemble v1.9.2

jpagh:
This PR fixes the import of `this_thread` to be from `docassemble.base.functions`. Docassemble v1.9.2 changed the handling of `this_thread` and instead of importing it into `docassemble.base.util`, now only references it by its full name, `docassemble.base.functions.this_thread`. This caused AssemblyLine's import of `this_thread` from `docassemble.base.util` to fail.

It might be worth considering a larger change to always directly reference `docassemble.base.functions.this_thread` like docassemble now does. Here is the relevant part of the commit,[ "changed way that this_thread is accessed to avoid any issues with it being redefined after being imported"](https://github.com/jhpyle/docassemble/commit/25240c65b3ff917481f08f122532fca31646b8b8).

Backwards compatibility note: `this_thread` has been defined in `docassemble.base.functions` for at least the past year, according to my quick search.

### DIFF
--- a/docassemble/AssemblyLine/al_general.py
+++ b/docassemble/AssemblyLine/al_general.py
@@ -1,6 +1,5 @@
 from copy import deepcopy
 from typing import Callable, Dict, List, Literal, Union, Optional, Any
-from docassemble.base.functions import this_thread
 from docassemble.base.util import (
     Address,
     as_datetime,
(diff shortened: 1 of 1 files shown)


nonprofittechy:
Makes sense, will ask @BryceStevenWilley to also review

PR #1015, opened 2026-02-22
make use of gathered attribute in ALDocumentBundle

### LINKED ISSUE #889: Enhance ALDocumentBundle by adding an affordance to handle dynamic item counts
ALDocumentBundle envisions a static cap on the number of documents you produce, but something, this number is dynamic.

For example: in Illinois, if you have multiple jobs, each job needs to recorded on a separate PDF document with a fixed format. The number of documents you produce is contingent on the number of jobs and there is no cap.

1. If you produce the bundle per-user, that's not a problem--just make the bundle an attribute of the user.
2. But if you want a single bundle to produce documents for an unbounded list, currently, you need to manually handle the list's contents and use a manual `reconsider` or perhaps `on update` block to handle refreshing the bundle's contents when the list changes.

I'm not sure yet what affordance makes sense. People can handle this manually with Python code, but a simpler API for common use cases would be helpful.

nonprofittechy:
Possibly fix #806, fix #889

I don't precisely remember why we didn't do that already--I think overly cautious about potential side effects when using ALDocument.

Clarification: bundles intended to use DAList's lazy automatic gathering must set both `auto_gather=True` and `gathered=False`. `auto_gather=True` is the switch that causes iteration to call the DAList gathering machinery, while `gathered=False` marks the bundle as incomplete. This gathering may resolve dependencies or ask questions before document filtering runs. It does not, by itself, mean that a custom code block will be used to populate `elements`.

Ordinary static bundles retain the default `auto_gather=False` and `gathered=True` behavior.

Validation update:

- Added real-class regressions for defined-false versus missing `gathered`, automatic gathering before child filtering, disabled nested bundles, and membership replacement reaching first PDF output.
- Added `test_bundle_gathering.yml` and an ALKiln feature covering the real resolver path. With `~/venv/bin/dainstall --server localhost`, the interview asks the automatic-gather dependency question before the results screen and reports: defined-false count 0, missing-marker count 1, automatic-gather count 2, and membership count 2.
- Browser-tested ProofOfDelivery, NameChangeChild, OrderOfProtection, VTFeeWaiver, VTFinancialDisclosure, VT813, MATCFinancialStatement, FinancialStatment, MotionToStayEviction, and EFSPIntegration on localhost. The named ILAO interviews and ProofOfDelivery reached their welcome screens after installing the shared ILAO package; the Vermont, MotionToStay, and EFSP interviews also loaded. MATCFinancialStatement reached its own interview error before the first question, and FinancialStatment reached a login gate. ProofOfDelivery required a disposable install copy because its current `pyproject.toml` has no `[project]` section.
- Static YAML checks found no bundle-related errors in the named examples. Existing unrelated lint findings remain in some downstream repositories (duplicate IDs, multiple mandatory blocks, missing IDs, or metadata completeness).

Screenshots and run notes are in the evidence gist linked from the review comment.

Opening as draft because it seems like a fundamental change that needs careful testing. But I don't see why it wouldn't solve what we want it to solve.

### DIFF
--- a/docassemble/AssemblyLine/al_document.py
+++ b/docassemble/AssemblyLine/al_document.py
@@ -1764,7 +1764,7 @@ class ALDocumentBundle(DAList):
         Returns:
             bool: True if there's at least one enabled document, otherwise False.
         """
-        return any(document.is_enabled(refresh=refresh) for document in self.elements)
+        return any(document.is_enabled(refresh=refresh) for document in self)
 
     def enabled_documents(self, refresh: bool = True) -> List[Any]:
         """
@@ -1776,11 +1776,7 @@ class ALDocumentBundle(DAList):
         Returns:
             List[Any]: List of enabled documents.
         """
-        return [
-            document
-            for document in self.elements
-            if document.is_enabled(refresh=refresh)
-        ]
+        return [document for document in self if document.is_enabled(refresh=refresh)]
 
     def as_flat_list(self, key: str = "final", refresh: bool = True) -> List[DAFile]:
         """


plocket (commented):
> Possibly fix https://github.com/SuffolkLITLab/docassemble-AssemblyLine/pull/806, fix https://github.com/SuffolkLITLab/docassemble-AssemblyLine/issues/889

It looks like #806 was already fixed. Did you mean a different issue?

Regarding the change, it seems like this may change the behavior of existing interviews depending on what architecture folks have set up. Are their arrangements of code that would break if docassemble tried to enforce a traditional `gather` loop?

nonprofittechy:
I ran a local runtime pass against PR #1015 head `ed4761a507c552e0102cc57f2e50dc88d505b779`.

Evidence/screenshots: [secret gist](https://gist.github.com/nonprofittechy/a974be4829023b311e3e3f19b8267d88)

Results:

- Installed the checked-out package on `http://localhost` with `~/venv/bin/dainstall --server localhost --debug .`.
- Installed a disposable `/tmp` interview with the same `dainstall` command and exercised the real `ALDocumentBundle` class in a browser.
- Static bundle behavior passed: enabled filtering returned 2 documents in order, and duplicate object identity was preserved.
- The stated manual-lazy edge remains confirmed: with `auto_gather=False, gathered=False`, `gathered` is still defined and iteration left the bundle empty. The new iterator does not, by itself, activate a builder for a stored `False` marker.
- The automatic path is behaviorally real: with `auto_gather=True, gathered=False`, `has_enabled_documents()` caused gathering before filtering and left `gathered=True`.
- Disabled nested-bundle short-circuit passed: the outer bundle returned `False` and its disabled auto-gather child remained un-gathered.
- The existing `test_aldocument.yml` browser walk completed 11 screens without a docassemble error page, covering PDF/DOCX bundle rendering, default/custom download-list HTML, and default/custom email-button rendering.

One limitation: `test_bundle_parity.yml` reached the normal PDF-generation wait screen but did not enable Next during the local wait window, so I am not counting parity/PDF output as a pass. The browser also reported an unrelated local docassemble chat-log callback error (`daChatLogCallback: Cannot read properties of null (reading 'length')`) before the PR-specific assertions; it did not produce an interview error page.

Conclusion: I found no regression in the ordinary static/nested/helper paths exercised here, but the runtime pass reinforces the existing concern that `gathered=False` does not implement the manual-lazy contract described in the PR. I would still request real-class/interview regressions distinguishing a missing marker from a defined `False`, plus an `auto_gather=True` question-order case and membership-change/output coverage before treating this as the complete fix for #889.

nonprofittechy:
## Additional regression and downstream validation

I added and ran the following coverage on the PR branch:

- Five real-class `ALDocumentBundle` regressions in `test_al_document.py`: a defined `gathered=False` marker remains defined and does not gather; a deleted marker is requested by iteration; `auto_gather=True` runs before child filtering for both helpers and forwards `refresh=False`; disabled nested bundles still short-circuit; and replacing membership before the first PDF output is reflected in that output.
- A real localhost interview/feature, installed with `~/venv/bin/dainstall --server localhost`, that exercises the resolver path. It pauses on an automatic-gather dependency question before the result screen and then reports `Defined False count: 0`, `Missing marker count: 1`, `Automatic gather count: 2`, `Automatic gather marker: True`, and `Membership count: 2`.
- `dayamlchecker` reports no issues for the new interview. The Python test file was syntax-checked; per the repository instructions I did not run the Python unit suite outside the full docassemble CI environment.

### Named downstream examples

I cloned the current default branches into `/tmp`, installed them on the localhost server where their package metadata allowed it, and inspected their `ALDocumentBundle` declarations and gathering flags.

| Repository | Result |
| --- | --- |
| `IllinoisLegalAidOnline/docassemble-ProofOfDelivery` | Welcome screen loaded. The repository’s current `pyproject.toml` has only a build-system block, so I used a disposable copy with that invalid file removed; the source itself was unchanged. |
| `IllinoisLegalAidOnline/docassemble-NameChangeChild` | Welcome screen loaded after installing shared `docassemble-ILAO`; explicit/manual bundle element replacement and list-comprehension construction were inspected. |
| `IllinoisLegalAidOnline/docassemble-OrderOfProtection` | Welcome screen loaded after installing shared `docassemble-ILAO`; static/nested bundle construction was inspected. |
| `LSVermont/docassemble-VTFeeWaiver` | Welcome screen loaded; static user/court bundles and manual list gathering were inspected. |
| `LSVermont/docassemble-VTFinancialDisclosure` | Welcome screen loaded; static user/court bundles and manual list gathering were inspected. |
| `LSVermont/docassemble-VT813` | Welcome screen loaded; static bundle construction and manual lists were inspected. |
| `SuffolkLITLab/docassemble-MATCFinancialStatement` | Static bundle declarations and YAML lint were inspected; the interview reached an interview error screen before the first question on this localhost image. |
| `SuffolkLITLab/docassemble-FinancialStatment` | Static source inspected; the interview reached its login gate. No `ALDocumentBundle` reference was found in this checkout. |
| `SuffolkLITLab/docassemble-MotionToStayEviction` | Welcome screen loaded; static/nested bundles and existing feature source were inspected. |
| `SuffolkLITLab/docassemble-EFSPIntegration` | Minimal and any-filing interviews loaded; the `auto_gather=True, gathered=False` court bundle path was inspected. Its package dependency metadata is incompatible with the localhost Python 3.14 image, so the runtime result is limited to the installed interview sources. |

The additional Vermont `VTFeeWaiverWithIncludeYMLFile` package was also installed and statically checked; it is an include-only package, not a standalone interview. YAML checking found no bundle-related errors in the named source. Existing unrelated downstream lint findings remain (duplicate IDs, multiple mandatory blocks, missing IDs, and metadata completeness).

The new screenshots are outside the repository in the evidence gist: https://gist.github.com/nonprofittechy/7e6140b3f96e0d2048c66c29ab3792eb

Conclusion: the clarified contract is now covered: automatic lazy bundle gathering requires both `auto_gather=True` and `gathered=False`, and it can ask a question before helper filtering. I found no regression in the ordinary static, nested, manual-membership, or helper paths exercised here. The existing limitation remains that this change does not itself invalidate cached output after a later membership edit; explicit rebuilding/reconsideration is still required for that scenario.

nonprofittechy:
This in some ways doesn't seem urgent, but also it's a long-standing correctness issue that has quite complex workarounds, like using in-place list comprehensions instead of more self-documenting conditional logic. I'm now quite confident that this is a safe change and will only come into effect when an author explicitly wants to change the bundle iteration behavior by defining two non-standard attributes in the .using() block. It will not change existing interview behavior.

PR #1032, opened 2026-03-21
feat: update language setting code to prioritize session language when available

jpagh:
As discussed with @nonprofittechy , this commit (9c6a08cd2033d241818634e5a376c11507807049) checks the `session_local` special variable for a language value before checking the interview-wide `al_user_language`.

This commit (73d3eaecfb3bcfacba45deaaef8024aef0577a0e) updates the event that the language dropdown calls so that the user's chosen language is stored in the `session_local`.

These together allows individual users to have a language selection that is unique to their session and doesn't change the interview-wide `al_user_language`.

If an interview is sent via URL then that still changes the interview-wide `al_user_language`.

### DIFF
--- a/docassemble/AssemblyLine/data/questions/al_language.yml
+++ b/docassemble/AssemblyLine/data/questions/al_language.yml
@@ -40,6 +40,14 @@ code: |
 ---
 event: al_change_language
 code: |
+  # Set browser-specific language for this interview session
   if "lang" in action_arguments():
     session_local.al_user_language = action_argument("lang")
     set_language(session_local.al_user_language)
+---
+event: al_get_language_list_change_language
+code: |
+  # Set "global" language for this interview session instead of user-specific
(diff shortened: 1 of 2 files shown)


nonprofittechy:
> I tested a bit, and I see the use case, but it does directly put that described usecase against users revisiting their interview after logging out (or more likely being logged out by their browser) and having their language revert to the default language of the interview. Having the initial URL param be "sticky" there is good IMO.

I think we can fix this by having the in-line version of the language switcher talk to a different event, which tends to appear on the first screen of multi-lingual interviews and reflects the primary user's language. We can also encourage authors to have an interview question to choose language and have that directly set the al_user_language.

This is already fine if the user visits the interview with a lang=es URL param set though, I believe?

BryceStevenWilley (approved):
Looks good to me!

PR #1033, opened 2026-03-26
Switched to PyPI trusted publishing

rajeswari1301:
Replaces the old PYPI_API_TOKEN publishing with PyPI's trusted publishing. This removes the need to store API secrets and generates temporary tokens automatically for each publish run.


BryceStevenWilley (approved):
Looks good to me!

PR #1039, opened 2026-04-03
Switch to only pyproject.toml

rajeswari1301:
Changes:
- Deleted setup.py, setup.cfg, and requirements.txt -since `pyproject.toml` already handled everything 
- Moved dev dependencies (mypy, type stubs, etc) into [dependency-groups] in pyproject.toml 
- Expanded MANIFEST.in 

Keeping everything in one file (pyproject.toml) is cleaner and sets us up to use `uv` in CI for faster installing.

Note: `pythontests` reusable workflow in ALActions currently installs from requirements.txt. That will need to be updated separately to install from [dependency-groups] instead.


BryceStevenWilley (approved):
Looks good to me!

PR #1046, opened 2026-04-30
Add alternative text to /s/signature (highest priority for accessible PDFs we send to MATC)

nonprofittechy:


BryceStevenWilley (commented):
Looks fine, but from what I can tell (checking with a screen reader best I can and through Acrobat) this alt text doesn't make it into the PDF?

nonprofittechy:
sounds like this might not do anything yet ~lol~ [for a DOCX file]. might still be worth merging, and I do not know yet if it has an impact on a signature field in a PDF. (I don't know if that gets alt_text or not)

Per Jonathan, we need an upstream update or subclass of the python-docx-template InlineImage request for this to make it into the DOCX file. I've opened that PR here: https://github.com/elapouya/python-docx-template/pull/636

PR #1049, opened 2026-05-13
Remove hardcoded CourtFormsOnline fallback; use proper global config appname, not app name

nonprofittechy:
We've kind of been polluting the brand "CourtFormsOnline" in the interview list on servers that aren't part of CourtFormsOnline because of two issues:

1) We're not using the `appname` config--we had a typo `app name` which isn't normally defined in a Docassemble config
2) The default when `appname` isn't defined was CourtFormsOnline. There's no good reason for that--we have 100% control over the servers where we want the interview list title to be CourtFormsOnline.

This PR just corrects the typo and removes the default CourtFormsOnline language, which is really a specific brand name for apps.suffolklitlab.org only.

### DIFF
--- a/docassemble/AssemblyLine/data/questions/al_document.yml
+++ b/docassemble/AssemblyLine/data/questions/al_document.yml
@@ -89,7 +89,7 @@ code: |
   # Used in the email subject line when email is triggered on download screen.
   # This defaults to the interview title. For example, if the
   # interview title is "Guardianship Helper", the email
-  # subject would be "Your Guardianship Helper document from CourtFormsOnline is ready".
+  # subject would be "Your Guardianship Helper document from docassemble is ready".
   # Customize this if you prefer something like "Your guardianship document is ready"
   x.document_label = str(
     all_variables(special="titles").get(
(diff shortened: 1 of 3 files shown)


docassemble/AssemblyLine/data/questions/al_language.yml
```diff
@@ -6,6 +6,10 @@ translations:
   - translation_es.xlsx
   - translation_ht.xlsx
   - translation_pt.xlsx
+  - translation_zh-CN.xlsx
+  - translation_zh-TW.xlsx
+  - translation_vi.xlsx
+  - translation_km.xlsx
```
BryceStevenWilley: None of these files exist in `sources` right now; I would guess you meant for this change to be in a different branch?

nonprofittechy: Yes, sorry, I carefully did a `git add` and then did `git commit -am `

BryceStevenWilley (approved):
Looks good to me, tried to think of a more generic term than docassemble but nothing really came to mind. Given it's the normal default for `appname` too, I think it's good.

PR #1051, opened 2026-05-13
Add alt_text attribute to signature's DAFile

### LINKED ISSUE #1050: Add alt text to signatures
See https://github.com/jhpyle/docassemble/issues/946 for the only way to do this in Docassemble for now - using validation code

nonprofittechy:
so it can be used in DOCX output in docassemble >= 1.9.10

See discussion of this attribute here: https://docassemble.org/docs/objects.html#DAFile.set_alt_text

And see issue opened on Docassemble, which is now resolved. See https://github.com/jhpyle/docassemble/issues/946

Fix #1050


BryceStevenWilley (approved):
Looks good to me! Noting that this only adds the alt text for the DOCX. Jonathan said it should also add it for PDFs if `tagged pdf: True` is there, but it doesn't (tried it, and the alt text doesn't appear anywhere in the PDF contents). So there's still a bit of work to be done there.

PR #1055, opened 2026-05-20
Add visually hidden document labels to buttons for accessibility

### LINKED ISSUE #1054: Download buttons are not accessible when navigated to with [TAB] key
The download/view links (as well as "Download all" and "Send") do not contain the name of the file that is being downloaded, viewed, or sent, making it inaccessible for a screen reader user.

Fixing this inspired some debate.

One "easy" fix would be to put the full name of each document in the button. This makes it much wider, and it also repeats the name of the document twice (for the "view" and "download" actions).

We also debated whether it was correct to use an anchor tag for this. HTML spec clearly says that links are to be used for downloads.

If we use an anchor tag, should we really style it like a button? This seems fine as long as we don't use role="button", as it's purely a visual change.

We might revisit the right fix, but for now we agreed to try a change that makes the links accessible without changing the visual style.

nonprofittechy:
Style is unchanged, but this allows screen reader users to be able to tell what button they are interacting with. Tested with Narrator and NVDA in Windows; this seems like a good fix.

Can revisit/discuss including full document text in visible label with more of our user base, but this should be compatible for now.

Fix #1054

Note: we should also update translations to match as this involves some Docassemble template edits, not just in the Python code.

I tested with the test_aldocument.yml file.

Edit: 94a5e35 moves the default text for send_button_html() to the right place; also has side effect of making this more consistently translatable (must have been missed in a prior change)

### DIFF
--- a/docassemble/AssemblyLine/al_document.py
+++ b/docassemble/AssemblyLine/al_document.py
@@ -2224,7 +2224,7 @@ class ALDocumentBundle(DAList):
                 download_doc.url_for(
                     attachment=True, display_filename=download_filename
                 ),
-                label=download_label,
+                label=f'{download_label} <span class="visually-hidden">{title}</span>',
                 icon=download_icon,
                 color="primary",
                 size="md",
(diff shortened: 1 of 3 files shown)


BryceStevenWilley (approved):
Tested with Voice Over on Mac and iPhone, works well on Mac, works on iPhone with a few quirks (it splits up the text inside each link, so that it reads "View [form title]", "View", "[form title]", "Download [form title]", "Download", "[form title]"). But each of those links still works, and the full context comes first, so I'm okay with it.

PR #1057, opened 2026-06-01
Make 'Email' input translatable

nonprofittechy:
This is a string we had overlooked from translatability (discovered by Matt)

### DIFF
--- a/docassemble/AssemblyLine/al_document.py
+++ b/docassemble/AssemblyLine/al_document.py
@@ -1602,6 +1602,9 @@ class ALDocumentBundle(DAList):
         email_input_label (DALazyTemplate): Template providing the label for the email input
             field in :meth:`send_button_html`. Defined generically in
             `ql_baseline.yml`; resolves to `"Email"` (or its translation).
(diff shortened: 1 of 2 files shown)


docassemble/AssemblyLine/al_document.py
```diff
@@ -2593,7 +2601,7 @@ def send_button_html(
   <div class="al_email_container">
   
     <span class="al_email_address {html_safe_str(self.instanceName)} container form-group row da-field-container da-field-container-datatype-email">
-      <label for="{al_email_input_id}" class="col-form-label da-form-label datext-right">Email</label>
+      <label for="{al_email_input_id}" class="col-form-label da-form-label datext-right">{email_label}</label>
       <input value="{user_info().email if user_logged_in() else ''}" alt="Email address for document" class="form-control" type="email" size="35" name="{al_email_input_id}" id="{al_email_input_id}">
```
BryceStevenWilley: Alt text here should probably be translated too, IMO `word()` would be fine if we don't want to add another orthogonal input

BryceStevenWilley (approved):
LGTM, should still address the alt text being translated too

PR #1060, opened 2026-06-04
Change `email_success` var name for doc bundles

### LINKED ISSUE #1059: email_success used both when sending emails to court and to user individually
AppealsMiscMotion uses / checkes `email_success` here: https://github.com/SuffolkLITLab/docassemble-AppealsMiscellaneousMotion/blob/ae266306d27b3064f12eb706e63ed7f620fa87dc/docassemble/AppealsMiscellaneousMotion/data/questions/appealsmiscellaneousmotion.yml#L448.

`email_success` is set both here: 

https://github.com/SuffolkLITLab/docassemble-AssemblyLine/blob/53391114b257833641f41417bf59eaa83a86c299/docassemble/AssemblyLine/data/questions/al_code.yml#L60

and here:

https://github.com/SuffolkLITLab/docassemble-AssemblyLine/blob/53391114b257833641f41417bf59eaa83a86c299/docassemble/AssemblyLine/data/questions/al_document.yml#L27

So we didn't attempt to send an email to the court because a user had already sent an email to themselves.

BryceStevenWilley:
Would conflict with the email to court features by redefining the same variable.

Fix #1059.

### DIFF
--- a/docassemble/AssemblyLine/data/questions/al_document.yml
+++ b/docassemble/AssemblyLine/data/questions/al_document.yml
@@ -24,7 +24,7 @@ code: |
       # We only send the PDF if "editable" checkbox wasn't checked, except when format limited to DOCX (only possible when len == 1)
       preferred_formats = ["pdf"]
 
-  email_success = x.send_email(
+  email_success_doc_bundle = x.send_email(
     to=action_argument("email"),
     template=(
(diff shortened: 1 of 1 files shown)


nonprofittechy (approved):
LGTM. Can't think of any external use of this variable

Issue #1056, opened 2026-05-20
Download screen and emailed document filename is not updated after changing the conditions that set it through the review screen


KelizaMcGuire:
**Describe the bug**
In the MLH Divorce interview, (once the user has reached the download screen for the first time) the Judgment of Divorce filename does not update after adding children or removing all children through the review screen.

The title of the document will be correct when opening it in a new tab from the download screen, and the document itself will be correct. However, the filename on the download screen and on the file that is emailed to the user will still be the original filename that was determined when the user first reached the download screen.

This bundle uses Bates page numbering, and it is the only conditionally named bundle in the interview.

See code setting filename and title from documents.yml in [repo](https://github.com/mplp/docassemble-MLHDivorceAndCustody):
```
objects:
  - judgment_and_addenda_bundle: ALDocumentBundle.using(elements=[
      judgment_ccmlh412,
      judgment_ccmlh413,
      judgment_addendum_personal_property_and_debt,
      judgment_addendum_vehicles_1_to_4,
      judgment_addendum_vehicles_5_to_8,
      judgment_addendum_vehicles_9_to_12,
      judgment_addendum_real_property_1_to_2,
      judgment_addendum_real_property_3_to_4,
      judgment_addendum_real_property_5_to_6,
      judgment_addendum_real_property_7_to_8,
      judgment_addendum_real_property_9_to_10,
      judgment_addendum_pensions,
      judgment_addendum_income_tax,
      judgment_addendum_parenting_time,
      judgment_overflow_addendum
      ], filename="Judgment of Divorce with Children CC-MLH-413" if there_are_marital_children else "Judgment of Divorce without Children CC-MLH-412", title="Judgment of Divorce with Children CC-MLH-413" if there_are_marital_children else "Judgment of Divorce without Children CC-MLH-412", enabled=True, add_page_numbers=True, page_number_area="BOTTOM_LEFT", page_number_digits=2, page_number_font_size=10, page_number_offset_horizontal=40, page_number_offset_vertical=25, page_number_prefix="Page ")
```

**To Reproduce**
1. Complete the Judgment of Divorce interview (or the complaint) with or without children.
2. On the download screen, note the name of the Judgment of Divorce (last file) and the title that appears when opened in browser.
3. Either add children (if you said no children originally) or remove all children (if you added them originally) through the review screen.
4. Click resume.
5. On the download screen, note the filename of the Judgment of Divorce and the title that appears when opened in browser.
[...]

PR #1064, opened 2026-06-09
[codex] Render translatable document titles at string boundaries

nonprofittechy:
## Summary

- render `ALDocument` titles to strings when `ALDocumentBundle.get_titles()` returns its documented `List[str]`
- convert upload document titles to strings before calling `.capitalize()`
- add regression coverage for template-backed document titles

## Why

Document and bundle titles can be supplied through docassemble template blocks (new default in https://github.com/SuffolkLITLab/docassemble-ALWeaver/pull/980) so they remain lazy and translate in the current interview language. Most rendering contexts already call `str()` implicitly, but `get_titles()` promises strings and `.capitalize()` requires a real string. Converting only at these boundaries preserves late translation and avoids freezing titles during object creation or background document caching.

## Validation

- `python -m pytest -q docassemble/AssemblyLine/test_al_document.py` (`8 passed`)
- `git diff --check`

### DIFF
--- a/docassemble/AssemblyLine/al_document.py
+++ b/docassemble/AssemblyLine/al_document.py
@@ -1936,7 +1936,7 @@ class ALDocumentBundle(DAList):
             if isinstance(document, ALDocumentBundle):
                 flat_list.extend(document.get_titles(key=key, refresh=refresh))
             else:
-                flat_list.append(document.title)
+                flat_list.append(str(document.title))
         return flat_list
 
     def as_pdf_list(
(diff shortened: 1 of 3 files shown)


BryceStevenWilley:
Is this supposed to be for #1056? or is it unrelated

nonprofittechy:
> Is this supposed to be for #1056? or is it unrelated

Unrelated. This is only handling DALazyTemplate as the .title attribute

BryceStevenWilley (approved):
Looks good to me!

PR #1072, opened 2026-06-25
[codex] Avoid duplicate BirthDate labels

nonprofittechy:
## Summary

Removes the custom `alMonthLabel`, `alDayLabel`, and `alYearLabel` entries from the baseline `BirthDate` field.

## Why

On the local docassemble 1.7.7 container, the integrated MATC 1A divorce package failed to assemble because the `BirthDate` field treated those label keys as duplicate field labels.

## Validation

Locally deployed AssemblyLine editable into the docassemble Docker container and ran the MATC1AJointPetition API driver end to end to the final download screen.

### DIFF
--- a/docassemble/AssemblyLine/data/questions/ql_baseline.yml
+++ b/docassemble/AssemblyLine/data/questions/ql_baseline.yml
@@ -446,9 +446,6 @@ question: |
 fields:
   - Birthdate: x.birthdate
     datatype: BirthDate
-    alMonthLabel: ${ word('Month') }
-    alDayLabel: ${ word('Day') }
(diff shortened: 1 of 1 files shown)


BryceStevenWilley:
> local docassemble 1.7.7 container

Why is it running on 1.7.7? All of the servers we run are on 1.9.8 at least (plus our patches).

I can also confirm that a server with the latest versions of both AssemblyLine (4.6.2) and ALToolbox (0.17.2) works fine on birthdate questions. Not trusting codex here.

nonprofittechy:
This is not a real PR, sorry. Running end to end tests on a stale local Docassemble.

PR #1073, opened 2026-06-26
Automatically prefix 'required' and 'maxlength' dictionaries

BryceStevenWilley:
Iterates through the `required` and `maxlength` dictionaries, so the caller doesn't have to prefix the label name with the full object name themselves.

The params now become:  `users[i].address_fields(required={'zip': True})`
Instead of the current: `users[i].address_fields(required={users[i].address.attr_name('zip'): True}`)


nonprofittechy (approved):
this is better, thanks.

PR #1075, opened 2026-06-30
[codex] Support docassemble 1.10 session APIs

### LINKED ISSUE #1074: Replace removed docassemble.webapp.db_object import for docassemble 1.10.0
## Summary

`docassemble/AssemblyLine/sessions.py` imports a private docassemble webapp helper that no longer exists in docassemble 1.10.0:

```py
from docassemble.webapp.db_object import init_sqlalchemy
...
db = init_sqlalchemy()
```

In current `jhpyle/docassemble` 1.10.0, `docassemble.webapp.db_object` is gone. The replacement low-level DB access helpers appear to live in `docassemble.webapp.db`, which exposes `get_session()` and `session_scope()` context managers.

## Why this breaks

This is an import-time failure: any code path that imports `docassemble.AssemblyLine.sessions` under docassemble 1.10.0 will fail with something like:

```text
ModuleNotFoundError: No module named 'docassemble.webapp.db_object'
```

The affected functions then use `db.connect()` to run raw SQL queries in:

* `get_saved_interview_list()`
* `find_matching_sessions()`
* `delete_interview_sessions()`

## Proposed fix

Replace the old private engine initialization with the 1.10.0 DB session helpers:

```py
from docassemble.webapp.db import get_session, session_scope
```

Then update the raw SQL call sites roughly as follows:

```py
# SELECT-style queries
with get_session() as session:
    rs = session.execute(get_sessions_query, parameters)
    for row in rs:
        sessions.append(dict(row._mapping))

# DELETE / write-style queries
with session_scope() as session:
    session.execute(delete_sessions_query, parameters)
```

This keeps the existing SQL mostly intact while avoiding the removed `init_sqlalchemy()` helper.

## Test suggestion

Add a simple import smoke test against docassemble 1.10.0:

```bash
python - <<'PY'
import docassemble.AssemblyLine.sessions
PY
```

A more complete regression test should exercise the three affected session-list/session-delete functions inside a docassemble app/database context.

## Notes

This is probably a relatively small code change, but it should be tested with a real initialized docassemble 1.10.0 database context because it changes from using a raw SQLAlchemy engine connection to docassemble's current session helpers.

nonprofittechy:
Closes #1074.

## What changed

- Use `docassemble.base.hooks.write_answer_json` on docassemble 1.10 while retaining a dynamic fallback to the legacy `server.write_answer_json` hook.
- Use `docassemble.webapp.db.get_session()` for every read query.
- Use transactional `docassemble.webapp.db.session_scope()` for every delete/update query, including the advisory-lock metadata upsert.
- Retain backwards compatibility with docassemble < 1.10 through the legacy engine's `connect()` and `begin()` context managers.

## Root cause

Docassemble 1.10 removed both `docassemble.webapp.db_object` and the `docassemble.base.functions.server` object. The original issue identified three database call sites, but `sessions.py` contained six database context blocks; this updates all of them so no legacy engine access remains on the v1.10 path.

## Impact

Importing `docassemble.AssemblyLine.sessions` works with docassemble 1.10 without dropping support for older installations. Read behavior is unchanged, while writes now use an explicit transaction in both compatibility paths.

## Validation

- `black --check docassemble/AssemblyLine`
- `mypy docassemble/AssemblyLine --config-file pyproject.toml`
- `python -m compileall -q docassemble/AssemblyLine`
- Built sdist and wheel with `python -m build`
- Import smoke test with installed docassemble 1.9.2 selected the legacy fallbacks
- Import smoke test with the docassemble 1.10.1 source selected `docassemble.webapp.db` and `docassemble.base.hooks`

Database operations themselves still require validation in a running docassemble app/database context.


nonprofittechy:
Going to try locally running v1.10 DA and run kiln against both this and the ALToolbox PR as a baseline.

PR #1076, opened 2026-07-01
Don't require all contact methods for al_cover_page

BryceStevenWilley:
Changes it from listing out all of the individual contact methods on separate table rows, requiring each to be present, to using `ALIndividual.contactMethods()`, which only shows those that are already defined. Also makes `user_needs_interpreter` optional.

Main reason I'm making this change: this document gets emailed to the court, but only after it gets e-filed. So it results in the user getting asked substantial questions like "do you need an interpreter" or "what's your phone number?" (which is designed to be optional) **after** efiling happens, IMO a bad and confusing user experience. Could (should?) consider removing some of the e-mail stuff entirely maybe, but this is a good immediate fix IMO.

DOCX diff:
From:

```
Preferred language:	{{ user_preferred_language if user_needs_interpreter else 'English' }}
Phone number:	{{ users[0].phone_number }}      
Mobile number:	{{ users[0].mobile_number }}
Email address:	{{ users[0].email }}
Contact Notes: 
(if any)
	{{ users[0].other_contact_method }}
```

```
Preferred language:	{{ user_preferred_language if showifdef(‘user_needs_interpreter’) else 'English' }}
Contact Information: 	{{ users[0].contact_methods() }}      
```


nonprofittechy:
Just wanted to make sure you knew about the Word Diff action-click on it and then Summary to see a pretty decent diff (it renders to Markdown as an intermediate step)

image

Where is this code still being used? Is this for older Appeals Court interviews that use email, or is it present in newer interviews?

BryceStevenWilley:
> Where is this code still being used? Is this for older Appeals Court interviews that use email, or is it present in newer interviews?

It's present on any of our interviews that call `sent_email_to_court` (currently in the MotionToStayEviction), which I thought was a thing that the courts still wanted so I had added it in the Small claims form, why I fixed it here. If we don't want it, I can remove it.

BryceStevenWilley:
Ended up removing `sent_email_to_court` from the MATC Small Claims, but leaving this PR open, since IMO the changes are useful. Unless we want to gut the whole thing, but I don't have a good pulse on what the Appeals court wants anymore.

nonprofittechy:
I do not think the trial court wants email notifications for filings. That is purely something the appeals court requested, but they have a fraction of the filing volume that the trial court has. But this is still worth fixing

nonprofittechy (approved):

PR #1077, opened 2026-07-01
Update ALKiln action to use docassemble image parameter

BryceStevenWilley:
Currently necessary because docassemble's latest version (1.10.0) changes a lot of stuff internally and our libraries (AssemblyLine, ALToolbox) don't work with it. We aren't targeting it for a bit, and don't want kiln tests to be failing that whole time.


BryceStevenWilley:
ALKiln test is currently failing on main, so I'm going to merge this. We can revert when if we want later, but I have been wanting to make this change for a while now since it's closer to testing what our servers currently run on.

PR #1080, opened 2026-07-19
issue with escaping dictionary keys in send_button_html

### LINKED ISSUE #1079: send_button_html (and related methods) do not properly escape DAObject intrinsic names that have a `'`, like dictionary keys
Issue: when you try to invoke send_button_html on an ALDocumentBundle that is in a dictionary, the on-screen rendering fails. Example: the object is named: `template_request.requestee_bundles['4167e36be7f04794ad30770e865afe68']`

Best resolution is likely to move away from a JavaScript string with hardcoded values, and instead use a data attribute and a click handler. We already can do this in the aldocument.js file we load

nonprofittechy:
Fix #1079 

A limited fix for issue escaping the intrinsic name of objects to construct the event name for send_button_html. This keeps the JavaScript link construction mostly as-is, except escaping for characters that would break the rendered HTML link.

Perhaps a better long-term fix would be to switch to using a JS listener and adding to the `data-` attribute, but this is more appropriate for now.

### DIFF
--- a/docassemble/AssemblyLine/al_document.py
+++ b/docassemble/AssemblyLine/al_document.py
@@ -1,6 +1,8 @@
-import re
-import os
+import json
 import mimetypes
+import os
+import re
+from html import escape
 from typing import Any, Dict, List, Literal, Union, Callable, Optional
 from docassemble.base.util import (
     Address,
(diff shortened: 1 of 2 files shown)


rajeswari1301 (approved):
Looks good to me. Ran the fix's tests locally (all passing, including the ones for the quote-escaping bug itself), and manually tested send_button_html() - worked cleanly

PR #1085, opened 2026-08-13
Speed up interview list query

BryceStevenWilley:
Now that we're getting Database performance data from ILAO, we can tell that the interview session query that happens in the interview_list.yml interview is taking up incredible percent of the total database cpu time (like 83% of the total database query time for today).

Running the existing query through an `EXPLAIN` a bit helped speed things up. By filtering the userdict entries to the desired user (using the userdictkey table, which has an index on `user_id`), it changed the estimated cost from 86522.63 to ~253, which should make things much faster. Tested on our dev server for a 5x speed up (0.95 load time for the whole page, to a 0.17), and I believe it should be even faster on servers with a lot of userdict rows.

I recommend y'all merge this, make a release and get it on ILAO prod ASAP to improve database performance (MLH doesn't seem to have problems with interview list interview, don't remember if they use it or not).

~Test failure is a mypy thing also happening on main, will try to fix it.~ Fixed!

### DIFF
--- a/docassemble/AssemblyLine/sessions.py
+++ b/docassemble/AssemblyLine/sessions.py
@@ -397,6 +397,12 @@ def get_saved_interview_list(
     filenames_to_exclude.extend([current_filename, filename_to_exclude])
 
     query_draft = """
+        WITH filtered_userdict AS (
+            SELECT userdict.indexno as indexno, userdict.key as key, userdict.dictionary as dictinory, userdict.modtime as modtime, userdict.user_id as user_id, userdict.filename as filename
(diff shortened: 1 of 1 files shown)


nonprofittechy (approved):
Great find!

rajeswari1301 (approved):
Looks great. One small thing to fix later, not blocking is the new query should use LEFT JOIN instead of JOIN for userdictkeys. Right now it'll hide about 2100 sessions from the admin "search all" feature. But doesn't affect regular users own lists.

PR #1086, opened 2026-08-13
Fix versions of docassemble to test with

BryceStevenWilley:
Newer versions of docassemble have extra typing information that is incompatible with our current code. Will have to address once we upgrade to 1.10.X.


BryceStevenWilley:
Merging, since main currently fails on mypy.

PR #1091, opened 2026-08-17
Add a subject so this has the right title for a screen reader

### LINKED ISSUE #1061: Terms of use text box not keyboard accessible
> The scrollable terms of use text box can be tabbed to in Chrome, but not all browsers support this by default. It needs "tabindex=0" and an accessible label for screen readers. https://dequeuniversity.com/rules/axe/4.11/scrollable-region-focusable?application=AxeChrome

> Add "tabindex=0" to the scrollable text box so it is focusable via keyboard in all browsers. Add a descriptive "aria-label" (such as "Terms of use") or use "aria-labelledby" to associate it with a visible heading so screen reader users understand what the region contains. Note: Firefox and Chrome already make scrollable regions keyboard accessible by default, with Safari being the remaining holdout (https://adrianroselli.com/2022/06/keyboard-only-scrolling-areas.html). Once Safari catches up, this tabindex may no longer be necessary.

(See A11Y-07)

nonprofittechy:
Fix #1061

Also updated the XLSX translation spreadsheet for Spanish; there wasn't a translation for any other languages yet


nonprofittechy:
Going to merge - we've visually previewed this, and it's a necessary accessibility change. No other substantive impact that needs testing.

PR #1092, opened 2026-08-17
Fix exhibit TOC page numbering

### LINKED ISSUE #884: Exhibit upload tool - TOC page numbering error
When using the exhibit upload tool to add multiple exhibits, the Table of Contents (TOC) displays wrong page numbers. Following the first exhibit, each subsequent exhibit page number shown in the TOC will be off by a higher number. 

Example: A user uses the tool to upload three separate 1-page documents. In the final exhibit document, the TOC will say that Exhibit A starts on page 3, which is correct. However, it will then say that Exhibit B starts on page 6 (actual start is page 5) and Exhibit C on page 9 (actual start is page 7).

This issue is mostly likely being caused by the logic in the default docx template for the TOC, shown below:
Screenshot 2024-09-18 at 8 55 28 PM (2)

Replacing `exhibit.start page + loop.index + table page length - 1` with `exhibit.start page + 1` (shown below) seems  to have resolved the issue for me when cover pages are included. However, this will need to be tested further.
Screenshot 2024-09-19 at 2 41 32 PM (2)

nonprofittechy:
Fixes #884.

## Root cause

`ALExhibitList._update_page_numbers()` already includes every preceding exhibit cover in each exhibit's `start_page`. The TOC template then added `loop.index`, counting those covers a second time. That produced the reported increasing error: `3, 6, 9` instead of the physical content pages `3, 5, 7`.

The no-cover expression also counted the default one-page TOC twice.

## What changed

- Add `ALExhibit.toc_page_number()` as the single, documented calculation for the first uploaded page.
- Have the DOCX TOC template call that method for cover and no-cover output.
- Add regression tests for:
  - one cover per exhibit
  - cover pages disabled
  - a multi-page TOC

## Before and after

Both screenshots come from the same local reproduction: three separate one-page exhibits assembled with a one-page TOC and one cover per exhibit.

| Before: incorrect cumulative offset | After: physical start pages |
| --- | --- |
| Before fix: TOC lists pages 3, 6, and 9 | After fix: TOC lists pages 3, 5, and 7 |

## Local validation

Installed each build into the localhost docassemble server with:

```bash
~/venv/bin/dainstall --server localhost .
```

Generated PDFs at `http://localhost` and extracted their TOC text:

- 3 one-page exhibits, covers enabled, 1-page TOC: 7-page PDF; starts `3, 5, 7`
- 3 one-page exhibits, covers disabled, 1-page TOC: 4-page PDF; starts `2, 3, 4`
- 28 one-page exhibits, covers enabled, 2-page TOC: 58-page PDF; starts `4, 6, …, 58`

Additional checks:

- `black --check docassemble/AssemblyLine/al_document.py docassemble/AssemblyLine/test_al_document.py`
- changed-module mypy check
- `git diff --check`
- DOCX ZIP integrity check
- `uv build --out-dir /tmp/issue884-dist`

The repository-wide mypy run still reports four unrelated existing errors in `language.py`, `al_general.py`, and `sessions.py`; the changed module passes independently.


nonprofittechy:
@rajeswari1301 this is not urgent at all (a very old correctness issue for a lightly used feature), but review if you can some day this week?

rajeswari1301 (approved):
The logic looks correct, and a test PDF with 3 exhibits and covers on came back with the right pages.

PR #1094, opened 2026-08-17
Hide literacy-only screen reader from tab order; it's redundant for low-vision users

### LINKED ISSUE #621: Remove "Listen to page" screen reader from tab order
I thought this was already in, but I guess it's not.

We should remove the entire widget from the tab order, and probably from accessibility tree in general, so it's not announced. Right now, it's still announced and is in the tab order. IMO we should do both.

nonprofittechy:
Fix #621


nonprofittechy:
Going to merge. This is a very straightforward fix.

PR #1095, opened 2026-08-18
Fix issue where you couldn't restart interview when in the middle of editing answers from review screen

### LINKED ISSUE #236: "Start over" on hamburger menu not working while on a screen triggered by "recompute" review screen logic
AB experienced an "Input not processed. Please try again" error when trying to select "Start over" from menu at top right.

She said when she tried to select it she was on a screen she had been taken to from the Review Screen.

Seems likely to be a higher-level Docassemble issue.

### LINKED ISSUE #1071: "Start over" on hamburger menu not working while on a screen triggered by "recompute" review screen logic
Edit something on the review screen and work through screens triggered by "recompute" logic on the review screen.

Example code showing the recompute logic I mean:
Image

While working through logic triggered by the "recompute", try to use the "start over" option on the hamburger menu. You get an "input not processed error".

[Original issue here](https://github.com/mplp/docassemble-MLHDivorceAndCustody/issues/236).

Notes from Suffolk standup in April: Does sound like it's inherent to this situation and the fact that Docassemble is mid-process, and this button requires starting a new process. AL folks thought they could maybe change how it confirms the action to use javascript instead and avoid this problem.

nonprofittechy:
Switch to using a JS prompt over url_ask

Tested end-end

fix #1071

### DIFF
--- a/docassemble/AssemblyLine/data/questions/al_visual.yml
+++ b/docassemble/AssemblyLine/data/questions/al_visual.yml
@@ -2,6 +2,10 @@
 include:
   - docassemble.ALToolbox:translation_strings.yml
 ---
+modules:
+  - json
+  - html
+---
 objects:
   - al_logo: DAStaticFile.using(filename="ma_logo.png", alt_text="")
 ---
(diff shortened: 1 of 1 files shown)


nonprofittechy:
@ekressmiller

PR #1097, opened 2026-08-25
Avoid duplicate session list query on interview_list page

rajeswari1301:
I was looking into the interview_list slow query and came across this - this page was calling `get_saved_interview_list( )` twice for the same data - once `inside session_list_html( )`, and again right after to check if there's a next page. This queries the list once and reuses it for both.


nonprofittechy:
What happens when you hit the "next" button on a paginated list: is the session list calculated fresh correctly?

rajeswari1301:
> What happens when you hit the "next" button on a paginated list: is the session list calculated fresh correctly?

Yes still works fine - that variable just lives for one page render, so it recalculates every time including after we hit next. only change is doing it once instead of twice

nonprofittechy (approved):
Thank you for spotting this! If it was doubling page load and people had a lot of interviews, it sounds like this should be a significant speedup. Much appreciated!

PR #1098, opened 2026-08-25
Skip exhibit pages with no valid files instead of crashing

rajeswari1301:
Fixes a bug where a user's document download kept failing. when every page for one file failed to load, it tried to build a PDF out of nothing and crashed the whole download instead of just skipping that file.

Three small fixes at each level where this could happen:
1. ALExhibit.as_pdf() - filters out missing or broken pages before building that exhibit's PDF. If nothing valid is left, logs it and returns None instead of crashing.
2. ALExhibitList.as_pdf() - now skips any exhibit that came back empty from step 1, and returns None if every exhibit ends up empty.
3. ALExhibitDocument.as_pdf() - handles step 2 returning None, so it doesn't crash trying to combine it with the table of contents.

None of these results get cached, so if something comes back on a retry (like it did in our actual case, which was a one-time blip), it's not stuck as skipped for the rest of the session.

Still need to test this against something close to the actual broken case before merging.

### DIFF
--- a/docassemble/AssemblyLine/al_document.py
+++ b/docassemble/AssemblyLine/al_document.py
@@ -1740,13 +1740,17 @@ class ALDocumentBundle(DAList):
             # In the case of no enabled files, avoid errors
             return None
         elif len(files) == 1:
-            # This case is simplest--we do not need to process the document at this level
             pdf = files[0].as_pdf(
                 key=key,
                 refresh=refresh,
                 pdfa=pdfa,
                 append_matching_suffix=append_matching_suffix,
             )
+            if pdf is None:
+                log(
+                    f"ALDocumentBundle.as_pdf(): '{files[0].title}' in bundle '{self.title}' has no valid PDF, skipping"
+                )
+                return None
             bundle_filename = f"{base_name(self.filename)}{append_suffix}.pdf"
             pdf.title = self.title
             pdf.filename = bundle_filename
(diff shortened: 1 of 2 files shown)


nonprofittechy (approved):
This looks good - wait to merge until you do your tests.

Main question is what happens if a `None` is able to escape--is it safe all the way up? I would try some adversarial bundles to test this.

Also: I see why this happened - we probably patched ALDocument but not ALExhibitList. Are there other descendants of ALDocument that we need to repair/guard in the same way? It would make sense to do those at the same time.

nonprofittechy:
Copilot's findings all also look worth addressing prior to merging, esp. updating the docstring. ideally we could also scan for now-unsafe use in our other repos

rajeswari1301:
> This looks good - wait to merge until you do your tests.
> 
> Main question is what happens if a `None` is able to escape--is it safe all the way up? I would try some adversarial bundles to test this.
> 
> Also: I see why this happened - we probably patched ALDocument but not ALExhibitList. Are there other descendants of ALDocument that we need to repair/guard in the same way? It would make sense to do those at the same time.

Good question, no not fully yet. Found a few more spots in ALDocumentBundle that assume as_pdf/as_docx never returns None, as_zip, `as_pdf_list`, `as_docx_list`, `get_cacheable_documents`, and a couple branches inside as_pdf itself. I'll add those to this PR too

PR #1100, opened 2026-08-27
Warn user when an uploaded exhibit could not be processed

rajeswari1301:
Follow up to #1098, which made broken exhibits fail silently instead of crashing the download, but it also means someone could end up missing a file from their packet and never know it. This adds a check for exhibits that finished uploading but came back with no valid pages, and shows a warning on the download screen when that happens.

It's built into `download_list_html()` itself so that way every interview using it gets the warning automatically. One thing worth knowing, this only catches broken exhibits, not other kinds of documents in a bundle. If a plain attachment fails for some other reason, this warning won't catch it.

### DIFF
--- a/docassemble/AssemblyLine/al_document.py
+++ b/docassemble/AssemblyLine/al_document.py
@@ -2203,6 +2203,30 @@ class ALDocumentBundle(DAList):
 
         return results, bundled_zip, bundled_pdf
 
+    def has_broken_documents(self) -> bool:
+        for document in self.enabled_documents():
+            if (
+                hasattr(document, "has_broken_exhibits")
+                and document.has_broken_exhibits()
+            ):
+                return True
+            if (
+                isinstance(document, ALDocumentBundle)
+                and document.has_broken_documents()
+            ):
+                return True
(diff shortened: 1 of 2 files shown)


nonprofittechy (approved):

docassemble/AssemblyLine/al_document.py
```diff
@@ -2203,6 +2203,44 @@ def get_cacheable_documents(
...
+
+        Returns:
+            str: The warning HTML, or an empty string if nothing is broken.
+        """
+        if not self.has_broken_documents():
+            return ""
+        return (
+            '<div class="alert alert-warning" role="alert">'
+            "One of your files did not upload correctly and it will not be included. "
+            "Please try uploading it again before you continue."
```
BryceStevenWilley, line 2240: I think we need to make a way to surface which document didn't upload correctly; some folks upload dozens of documents as exhibits, without specific guidance here re-uploading everything could take them a significant amount more time (not to mention could re-trigger the root issue if it's related to the number of exhibits that uploaded at the same time).

PR #1101, opened 2026-08-30
Name the broken exhibit in the warning message

rajeswari1301:
Follow-up to #1100, the warning used to just say "one of your files", without saying which one, and asks the user to just try again, which isn't great when someone uploads multiple exhibits.

Now it displays the name of a broken exhibit(s) by title, and only tells you to re-upload those, not everything.


BryceStevenWilley (approved):
Looks good to me! Couldn't end up recreating the issue locally, but can confirm at least that it works fine as normal.

Issue #1102, opened 2026-08-31
Move interview list into a flask endpoint


nonprofittechy:
Goals:

1. Allow us to optimize page load speed avoid reload on undefined variable resolution
2. Let us avoid polluting the database with sessions that don't need to be stored. we do some management already by minimizing the session lifetime, but there's still a reasonably large amount of churn

PR #1104, opened 2026-09-08
Add Flask interview list endpoint

rajeswari1301:
Replaced the redundant session query in interview_list.yml, which was executing 2–3 times per page load and now runs only once, tested it at  50 - 2375 saved sessions - and it was about about 2-3 times faster each time, getting better the bigger the list gets.

It supports all the same features - viewing, searching, renaming, deleting, copying, and paging through results and pulls its settings from the same config the old page already uses.

#1102

### DIFF
--- a/docassemble/AssemblyLine/interview_list_endpoint.py
+++ b/docassemble/AssemblyLine/interview_list_endpoint.py
@@ -1,0 +1,395 @@
+# pre-load
+import time
+
+from flask import request, redirect, url_for, flash, render_template_string
+from flask_login import login_required, current_user
+from flask_wtf.csrf import generate_csrf
+from markupsafe import escape
+
+from docassemble.webapp.app_object import app
+from docassemble.webapp.server import user_interviews
+from docassemble.base.functions import this_thread, interview_url, log, get_config
+from docassemble.AssemblyLine.sessions import (
+    get_saved_interview_list,
+    find_matching_sessions,
+    delete_interview_sessions,
+    rename_interview_answers,
+    save_interview_answers,
+    nice_interview_title,
+    nice_interview_subtitle,
+    radial_progress,
+    local_date,
+    config_with_language_fallback,
+    get_combined_filename_list,
+)
+
+PAGE_SIZE = 20
+
+PAGE_TEMPLATE = """
+{%- extends 'flask_user/public_base.html' %}
+{%- block content %}
+<link rel="stylesheet" href="/packagestatic/docassemble.AssemblyLine/interview_list.css">
+<style>
+.al-icon-btn {
+    background: none;
+    border: none;
+    padding: 0;
+    cursor: pointer;
+    color: #0d6efd;
+    text-decoration: none;
+}
+.al-icon-btn:hover, .al-icon-btn:focus {
+    text-decoration: underline;
+}
+.al-icon-btn i {
+    color: #0d6efd;
+}
+.al-icon-btn.al-danger, .al-icon-btn.al-danger i {
+    color: #dc3545;
+}
+.al-icon-btn span {
+    font-size: 0.85rem;
+}
+</style>
+{{ content|safe }}
+{%- endblock %}
+"""
+
+DEFAULT_EXCLUDED_FILENAMES = [
+    "docassemble.ALDashboard",
+    "docassemble.AssemblyLine:data/questions/al_saved_sessions_store.yml",
+    "docassemble.AssemblyLine:data/questions/interview_list.yml",
+]
+
+
+def _list_config():
+    """Pulls the same "assembly line: interview list" settings interview_list.yml
+    reads, so this page stays in sync with however each site has it configured."""
+    interview_list_config = get_config("assembly line", {}).get("interview list", {})
+    exclude_filenames = interview_list_config.get(
+        "exclude from interview list",
+        get_config("assembly line", {}).get(
+            "exclude from interview list", DEFAULT_EXCLUDED_FILENAMES
+        ),
+    )
+    return {
+        "page_heading": config_with_language_fallback("page question", "interview page heading") or "In progress forms",
+        "page_intro": config_with_language_fallback("interview page pre"),
+        "new_form_label": config_with_language_fallback("new form label") or "Start a new form",
+        "new_form_url": config_with_language_fallback("new form url", "app homepage"),
+        "enable_answer_sets": bool(get_config("assembly line", {}).get("enable answer sets")),
+        "exclude_filenames": exclude_filenames,
+    }
+
+
+def _set_current_info():
+    # rename_interview_answers/save_interview_answers/user_interviews all read
+    # this_thread.current_info for the acting user, which usually is populated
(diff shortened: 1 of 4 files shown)


docassemble/AssemblyLine/interview_list_endpoint.py
```diff
@@ -0,0 +1,443 @@
...
+.al-icon-btn i {
+    color: #0d6efd;
+}
+.al-icon-btn.al-danger, .al-icon-btn.al-danger i {
+    color: #dc3545;
+}
+.al-icon-btn span {
+    font-size: 0.85rem;
+}
+</style>
```
nonprofittechy: Maybe put all the CSS into the .css file instead of some

nonprofittechy (commented):
Minor nits. Biggest question to consider: should this definitely use HTML encoded into the file? An alternative is to use a traditional external HTML template. You'd need to load it in as a string though given that Docassemble only looks for Flask templates in one hardcoded folder.

docassemble/AssemblyLine/interview_list_endpoint.py
nonprofittechy: Also we shouldn't hardcode the package name here (line 31) if we can avoid it

rajeswari1301:
> Minor nits. Biggest question to consider: should this definitely use HTML encoded into the file? An alternative is to use a traditional external HTML template. You'd need to load it in as a string though given that Docassemble only looks for Flask templates in one hardcoded folder.

Thank you for the review! I fixed the CSS and package name. 

Right now, the page wrapper currently uses a separate file. Registering the package's template folder directly with flask would be the fix rather than another workaround, but that would mean updating every link on the page and handling the check that stops this route from loading twice

nonprofittechy:
> > Minor nits. Biggest question to consider: should this definitely use HTML encoded into the file? An alternative is to use a traditional external HTML template. You'd need to load it in as a string though given that Docassemble only looks for Flask templates in one hardcoded folder.
> 
> Thank you for the review! I fixed the CSS and package name.
> 
> Right now, the page wrapper currently uses a separate file. Registering the package's template folder directly with flask would be the fix rather than another workaround, but that would mean updating every link on the page and handling the check that stops this route from loading twice

We cannot register the package's template folder. But we can put templates in the data/sources or data/templates folder, and read them in as a string using a normal filesystem operation.

I think that would be more readable and maintainable than hardcoding the template in the Python module. What do you think?

nonprofittechy:
Expanding on the suggestion to move the HTML out of `interview_list_endpoint.py`, I think we can make the separation fairly complete without registering a new Flask template directory.

### Proposed implementation plan

1. **Keep the current filesystem-loading approach.** Put the full page template in a package data directory (for example `data/templates/al_interview_list_page.html` or another directory we know is packaged), load it with `Path(...).read_text()`, and continue to use `render_template_string()`. This avoids depending on Flask template-path registration.

2. **Move essentially all page markup into the Jinja template.** The template should own the heading, new-form button, intro text, search/filter form, saved-session table, rename/copy/delete controls, pagination, and delete-all form. The current Python helpers that return HTML strings (`_toggleable_action_icon()`, `_toggleable_action_form()`, `_render_session_rows()`) can then go away.

3. **Use Jinja macros for repeated UI pieces.** Rename/copy controls are good candidates for macros. This keeps the template readable while still making the presentation reusable.

4. **Have the Flask route prepare data, not HTML.** The GET route should be responsible for request parsing, fetching/filtering sessions, config, CSRF token generation, filename choices, and pagination state, then pass those values into the template. Something along the lines of:

   ```python
   return render_template_string(
       PAGE_TEMPLATE,
       sessions=sessions,
       filename_options=filename_options,
       cfg=cfg,
       csrf_token=csrf_token,
       page=page,
       keyword=keyword,
       limit_filename=limit_filename,
       page_size=PAGE_SIZE,
       package_name=_package_name(),
   )
   ```

5. **Optionally prepare a small view-model for each session in Python.** For example, Python can calculate `title`, `subtitle`, interview URL, formatted modification date/time, and progress markup. That keeps AssemblyLine-specific helper calls out of the template while still leaving the actual HTML structure in Jinja.

6. **Let Jinja handle normal escaping.** Right now a large generated HTML blob ultimately gets inserted with `{{ content|safe }}`. Moving the markup into Jinja lets normal values use Jinja autoescaping and limits `|safe` to values that are intentionally trusted HTML (for example, if `radial_progress()` returns markup). That gives us a safer default for future edits.

7. **Keep CSS in `interview_list.css`, and ideally move inline behavior/style out over time.** The current inline `style=`, `onclick=`, and `onsubmit=` attributes can work initially, but once the HTML is in a real template they will also be much easier to migrate to CSS/JS if desired.

8. **Leave route registration and endpoint names alone.** The existing `if "al_interview_list" not in app.view_functions:` guard can stay exactly where it is, and the template can call `url_for()` directly, so moving the markup should not require changing the route-loading strategy or URLs.

9. **Add/adjust tests around rendered behavior rather than generated strings.** At minimum I would cover: empty list, normal rows, answer-set controls on/off, search/filter state preserved, pagination links preserving query parameters, and escaping of user-controlled titles/names.

The end state I have in mind is: **Python handles auth/request/data/mutations; Jinja handles layout and conditional presentation; CSS handles styling.** That should make this endpoint considerably easier to review and maintain while keeping the current docassemble/Flask constraints intact.

rajeswari1301:
> > > Minor nits. Biggest question to consider: should this definitely use HTML encoded into the file? An alternative is to use a traditional external HTML template. You'd need to load it in as a string though given that Docassemble only looks for Flask templates in one hardcoded folder.
> > 
> > 
> > Thank you for the review! I fixed the CSS and package name.
> > Right now, the page wrapper currently uses a separate file. Registering the package's template folder directly with flask would be the fix rather than another workaround, but that would mean updating every link on the page and handling the check that stops this route from loading twice
> 
> We cannot register the package's template folder. But we can put templates in the data/sources or data/templates folder, and read them in as a string using a normal filesystem operation.
> 
> I think that would be more readable and maintainable than hardcoding the template in the Python module. What do you think?

Thanks for the thorough review and suggestion! I checked the existing conventions in this repo, and since ALDashboard already stores full-page templates in data/templates, I moved the HTML out of the python module and into 'data/templates/al_interview_list_page.html'. I also refactored the helper functions, updated the tests to cover the rendered template, it's ready for another look whenever you have a chance. Thanks again!

nonprofittechy (approved):
Good to merge. Feel free to make further changes to my revisions

BryceStevenWilley (approved):
Looks good to me. Some slight differences in style between the existing page and this page; it's fine with me, but we'll have to see if partners want to tweak it at all

rajeswari1301:
> Good to merge. Feel free to make further changes to my revisions

Thanks for taking a look and for the updates. I made one small follow-up to the filename lookup so it uses `userdictkeys` for ownership, and I also made the search dropdown use the existing filename exclusions. I added tests for both and everything is passing. I’ll go ahead and merge this now.

PR #1105, opened 2026-09-14
Skip pages that pass validation but fail to read during concatenation

rajeswari1301:
There were few "no valid files to concatenate" crashes even after the last fix. The page already passed the check, but the file still failed when the code went to use it, probably a storage write that wasn't fully finished. That specific step was not handled, so it crashed the whole document instead of just skipping that one piece.

This catches the failure right where it happens and skips it instead of crashing, across the exhibit, the exhibit list, and the combined exhibit doc.

Also added: when an exhibit fails this way, it now marks itself as broken so the existing warning message shows it to the user, instead of it just quietly disappearing from the download. Updated the warning text too, since it used to say the file didn't upload correctly, but the upload was fine, it's the processing afterward that failed.

It stops the crash and tells the user something went wrong.


nonprofittechy:
Not sure about this one, since we're silently skipping a file that the user uploaded, right?

rajeswari1301:
> Not sure about this one, since we're silently skipping a file that the user uploaded, right?

Yes that's correct, that was happening, sorry I missed it the first time. I fixed it now, the exhibit marks itself broken when this happens, so the warning shows it instead of it silently disappearing. My guess is it may be tied to the image conversion itself, I saw a few JPG conversion failures right before the crash in at least one session, though I'm not sure yet if that's the actual cause or something else going on, like a timing or file-handling issue.

nonprofittechy:
great, with that clarification + update will leave for @BryceStevenWilley to finish review unless he prefers me to finish

BryceStevenWilley (approved):
LGTM!

